PR #19, opened 2019-05-20
Create logfiles in $HOME/.raiden."

nlsdfnbch:
Instead of creating log files where the file is created, we now default to:
```
$HOME/.raiden/scenario-player/<scenario-name>/<log>.log
```

Which puts all generated log files into the same location.

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -75,7 +75,7 @@ def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, m
     else:
         scenario_basename = basename(scenario_file.name)
         log_file_name = (
-            f"{os.environ['HOME']}/.raiden/scenario-player/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
+            f"{data_path}/scenario-player/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
         )
     click.secho(f"Writing log to {log_file_name}", fg="yellow")
     configure_logging(
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -75,7 +75,7 @@ def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, m
     else:
         scenario_basename = basename(scenario_file.name)
         log_file_name = (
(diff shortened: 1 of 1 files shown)


scenario_player/main.py
```diff
@@ -75,7 +75,7 @@ def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, m
     else:
         scenario_basename = basename(scenario_file.name)
         log_file_name = (
-            f"scenario-player_{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
+            f"{os.environ['HOME']}/.raiden/scenario-player/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
```
ulope: This should re-use the `data_path` that is passed in from the command line:

https://github.com/raiden-network/scenario-player/blob/a70c56970b8ea311cdec2ca3f757bcc6aaab132e/scenario_player/main.py#L115

https://github.com/raiden-network/scenario-player/blob/a70c56970b8ea311cdec2ca3f757bcc6aaab132e/scenario_player/runner.py#L66

Also it would make sense to include the run number in the log name (as it's done for the node logs).

nlsdfnbch: huh. I missed that variable. will do!

nlsdfnbch: As for the run number, I think that's determined somewhere deeper in the code, so I'm leaving it our for now.

ulope: Yeah that's true. Since the debug log file uses the `RotatingFileHandler` we should get separate files for every run anyway.

scenario_player/main.py
```diff
@@ -75,7 +75,7 @@ def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, m
     else:
         scenario_basename = basename(scenario_file.name)
         log_file_name = (
-            f"{os.environ['HOME']}/.raiden/scenario-player/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
+            f"{data_path}/scenario-player/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
```
ulope: `data_path already` contains the `/scenario-player/` prefix.

PR #21, opened 2019-05-20
Add rc post hook to log packer

nlsdfnbch:


scenario_player/main.py
```diff
@@ -180,5 +182,121 @@ def reclaim_eth(obj, min_age):
...
+    )
+
+    token = resp.json()['data']['authToken']
+    user_id = resp.json()['data']['userId']
+    headers = {
+        'X-Auth-Token': token,
+        'X-User-Id': user_id,
+    }
+
+    with fpath.open('rb') as f:
```
ulope: IMO those files should rather be uploaded to something intended to store files (e.g. a DO spaces bucket) and then a link posted to RC.

One reason is that it will become pretty difficult to find older logs as time goes by and additionally in our current setup all attachments posted to RC are stored inside the mongodb, which makes backups annoying as they keep growing with attachements.

setup.py
```diff
@@ -13,6 +13,7 @@
     entry_points={
         'console_scripts': [
             'scenario-player=scenario_player.__main__:main',
+            'sp-log-packer=scenario_player.__main__:pack_logs',
```
ulope: Since you already defined `pack-logs` as a subcommand on the `main` group this isn't really necessary.

scenario_player/main.py
```diff
@@ -180,5 +182,121 @@ def reclaim_eth(obj, min_age):
...
+        raise RuntimeError(f"{raiden_dir} does not exist!")
+
+    target_dir = Path(target_dir)
+    target_dir.mkdir(exist_ok=True)
+
+    files = set()
+
+    scenario_names = scenario_name or []
+
+    for scenario_name in (scenario_names or []):
```
ulope: This loop will not execute if no scenario names have been passed.

scenario_player/main.py
```diff
@@ -180,5 +182,121 @@ def reclaim_eth(obj, min_age):
...
+    '--pack-n-latest', default=1,
+    help='Specify the max num of log history you would like to pack. Defaults to 1.'
+         'Specifying 0 will pack all available logs for a scenario.',
+)
+@click.option(
+    '--raiden-dir', default=os.environ.get('HOME', '.') + '/.raiden',
+    help="Path to the raiden meta data dir. Defaults to ~/.raiden.",
+)
+@click.option('--post-to-rocket', default=True)
+def pack_logs(post_to_rocket, raiden_dir, pack_n_latest, scenario_name, target_dir):
```
ulope: Nitpick: since `multiple=True` args always get passed as lists it would make sense to call the parameter `scenario_names`. You can to this in click by adding another positional argument that **doesn't** start with a dash.
E.g.:
```python
@click.option("--some-arg", "some_args", multiple=True)
```

scenario_player/main.py
nlsdfnbch: Agreed. For this iteration, this will have to do though.  The notification / reporting services is the first of the things I will add after the nightlies run without throwing hissiy fits. They will then only send download URLs.

setup.py
nlsdfnbch: I...did? Good to know! :smile:

scenario_player/main.py
nlsdfnbch: aha! Cool, thanks!

setup.py
ulope: `main()` is defined as a `@click.group()` that means any command that get's added via `@main.command()` will become a subcommand.

PR #22, opened 2019-05-20
Add simple circleci config.

nlsdfnbch:


nlsdfnbch:
Added lint step as well. Adding `tests/unit-tests` folders as well, so the `unit-test` step should be a no-op (since it's empty) -otherwise I believe it'll throw an error because the folder doesnt exist.

PR #23, opened 2019-05-21
Refac CLI options and commands.

nlsdfnbch:


scenario_player/main.py
```diff
@@ -53,42 +88,38 @@
...
+    chain_rpc_urls = parse_chain_rpc_urls(chains)
+
+    if ctx.invoked_subcommand:
+        ctx.obj = dict(
+            chain_rpc_urls=chain_rpc_urls,
+            data_path=data_path,
+        )
+
+
+@main.command(name="run")
```
ulope: Just FYI: if the command name is identical to the callable you don't need to provide the `name` argument.

scenario_player/main.py
```diff
@@ -42,9 +42,44 @@
 TRANSACTION_DEFAULTS["gas"] = lambda web3, tx: web3.eth.estimateGas(tx) * 2
 
 
+def construct_log_file_name(scenario_fpath: Path=None)-> str:
+    if scenario_fpath:
+        scenario_basename = basename(scenario_file.name)
+        return f"{data_path}/scenarios/{scenario_basename}/{scenario_basename}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
+    return f"{data_path}/scenario-player-{ctx.invoked_subcommand}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
```
ulope: `ctx` is not defined here

scenario_player/main.py
```diff
@@ -169,71 +191,75 @@ def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, m
...
-def pack_logs(post_to_rocket, raiden_dir, pack_n_latest, scenario_names, target_dir):
-    raiden_dir = Path(raiden_dir)
-    if not raiden_dir.exists():
-        raise RuntimeError(f"{raiden_dir} does not exist!")
+@click.argument("scenario-file", type=click.File(), required=False)
+@click.pass_context
+def pack_logs(ctx, scenario_file, post_to_rocket, pack_n_latest, target_dir):
+    data_path = ctx.obj['data_path']
+
+    scenario_name = Path(scenario_file).stem
```
ulope: This won't work `scenario_file` is a `click.File()` argument type which generates a File-like object. 
You can get at the path via the `.name` attribute but the better option would be to use `click.Path(exists=True, file_okay=True, dir_okay=False)` which directly returns the path as a string.

(Looks like the way we use `scenario_file` in what is now `run()` could also be changed to `Path` since we only ever use the `.name` attribute)

scenario_player/main.py
```diff
@@ -242,29 +268,25 @@ def pack_logs(post_to_rocket, raiden_dir, pack_n_latest, scenario_names, target_
...
-        try:
-            files.add(scenario_logs[i])
-        except IndexError:
-            # We ran out of scenario logs to add before reaching the requested number of n latest logs.
-            print(
-                f'Only packed {i} logs of requested latest {n} '
-                f'- no more logs found for {scenario_name}!',
-            )
-            break
+    if num_of_packable_iterations < n:
```
ulope: I don't get this logic. 

`num_of_packable_iterations` is a `range` (why?) which isn't comparable to `int`. 
Additionally it's initialized from `n or len(scenario_logs)` but if `n` can ever be Falsey this condition doesn't makes much sense even if it were possible to compare `range` and `int`

scenario_player/main.py
```diff
@@ -242,29 +268,25 @@ def pack_logs(post_to_rocket, raiden_dir, pack_n_latest, scenario_names, target_
...
+        # We ran out of scenario logs to add before reaching the requested number of n latest logs.
+        print(
+            f'Only packing {num_of_packable_iterations} logs of requested latest {n} '
+            f'- no more logs found for {scenario_name}!',
+        )
 
-    return files
+    return history[:num_of_packable_iterations]
```
ulope: Depending how (or if at all) log rotation / purging is done, it could with time become expensive to build a list of all logs and then throwing most of it away. 

Can be improved alter of course.

scenario_player/utils.py
```diff
@@ -1,5 +1,6 @@
 import json
 import os
+import pathlib
```
ulope: `Path` is already imported below

scenario_player/main.py
```diff
@@ -42,9 +42,46 @@
 TRANSACTION_DEFAULTS["gas"] = lambda web3, tx: web3.eth.estimateGas(tx) * 2
 
 
+def construct_log_file_name(sub_command, data_path, scenario_fpath: Path=None)-> str:
+    file_name = f"scenario-player-{sub_command}_{datetime.now():%Y-%m-%dT%H:%M:%S}.log"
+    directory = data_path
+    if scenario_fpath:
+        directory.joinpath(scenario_fpath.stem)
```
ulope: `Path.joinpath()` doesn't modify the Path in place, it returns a new one instead

scenario_player/main.py
```diff
@@ -53,42 +90,40 @@
...
+@click.password_option("--password", envvar="ACCOUNT_PASSWORD", required=True)
 @click.option("--auth", default="")
 @click.option("--mailgun-api-key")
-@click.argument("scenario-file", type=click.File(), required=False)
 @click.pass_context
-def main(ctx, scenario_file, keystore_file, password, chains, data_path, auth, mailgun_api_key):
-    gevent.get_hub().exception_stream = DummyStream()
+def run(ctx, mailgun_api_key, auth, password, keystore_file, scenario_file):
+    scenario_file = Path(scenario_file.name).absolute()
+    print(scenario_file)
```
ulope: Left over print

Issue #30, opened 2019-05-22
Use fresh accounts without renaming scenario file


czepluch:
Right now, if you run into problems with a corrupt node when running in `managed` mode, you can only create new nodes by renaming a scenario. There should be a way to generate new addresses used for the nodes in the scenarios, without having to rename the scenario file.

czepluch:
@nlsdfnbch can we please prioritize this, since the scenarios will always fail until we have some mechanism for this.

nlsdfnbch:
Must node names be identical within a scenario run? i.e. if I stop and start nodes as  a task, should those still have the same node name? @czepluch , @ulope

czepluch:
What do you mean by node name here? The address?

ulope:
Yes, they absolutely must be the within a single run.

As a quick fix I’d suggest to just add the run number into the seed of the private key. We can then work on a better solution once you’re back from vacation.

PR #35, opened 2019-05-28
Force Node refresh on new day.

nlsdfnbch:
Semi-ideal way to fix issue #30.


ulope:
That's effectively the same as deleting the node db.
But since the address will stay the same the state on-chain and in the MS and PFS will still be that of the previous run.

This would most likely cause hard to track down issues.

PR #38, opened 2019-05-29
#30 - Encode run number into private key.

nlsdfnbch:


scenario_player/node_support.py
```diff
@@ -176,7 +176,7 @@ def __init__(self, runner: ScenarioRunner, index: int, raiden_version, options:
         self._index = index
         self._raiden_version = raiden_version
         self._options = options
-        self._datadir = runner.data_path.joinpath(f"node_{index:03d}")
+        self._datadir = runner.data_path.joinpath(f"node_{self._runner.run_number}_{index:03d}")
```
ulope: That’s not absolutely necessary since raiden itself partitions it’s database location inside the datadir by keystone address and network id but probably makes looking at the results easier.

PR #50, opened 2019-06-13
Add option to check that no ClaimReward event was triggered

palango:
@czepluch Seems like we need this for MS3.

### DIFF
--- a/scenario_player/tasks/blockchain.py
+++ b/scenario_player/tasks/blockchain.py
@@ -217,12 +217,11 @@ class AssertMSClaimTask(Task):
         events = [e for e in events if match_event(e)]
         log.info("Matching events", events=events)
 
-        should_fail = self._config.get("should_fail", False) == True
+        must_claim = self._config.get("must_claim", True)
         found_events = len(events) > 0
 
         # Raise exception when no event was found
-        if not should_fail and not found_events:
+        if must_claim and not found_events:
             raise ScenarioAssertionError("No RewardClaimed event found for this channel.")
-
-        if should_fail and found_events:
+        elif not must_claim and found_events:
             raise ScenarioAssertionError("Unexpected RewardClaimed event found for this channel.")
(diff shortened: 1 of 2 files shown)


scenario_player/tasks/blockchain.py
```diff
@@ -217,6 +217,12 @@ def match_event(event: Dict):
         events = [e for e in events if match_event(e)]
         log.info("Matching events", events=events)
 
+        should_fail = self._config.get("should_fail", False) == True
+        found_events = len(events) > 0
+
         # Raise exception when no event was found
-        if len(events) == 0:
+        if not should_fail and not found_events:
```
rakanalh: Assuming that `should_fail` from config is set to `True`, this line will evaulate the first condition to False. 
Shouldn't it be like this?

```suggestion
        if should_fail and not found_events:
```

palango: No, the scenario player logic is, throw an exception if something goes wrong, otherwise success is assumed. So here we don't want the task to fail, but there aren't any events found, so we throw a `ScenarioAssertionError`.

ulope: This is a bit of a personal preference thing, but I always find inverted logic difficult to comprehend. 
I'd suggest to invert the parameter name. e.g. `must_claim` (with default `True`) then the condition would become the (IMO) much easier to parse `must_claim and not found_events`

scenario_player/tasks/blockchain.py
```diff
@@ -217,6 +217,12 @@ def match_event(event: Dict):
...
             raise ScenarioAssertionError("No RewardClaimed event found for this channel.")
+
+        if should_fail and found_events:
```
ulope: Nitpick: this could be an `elif`

scenario_player/tasks/blockchain.py
```diff
@@ -217,6 +217,12 @@ def match_event(event: Dict):
         events = [e for e in events if match_event(e)]
         log.info("Matching events", events=events)
 
+        should_fail = self._config.get("should_fail", False) == True
```
ulope: Nitpick `True == True`: `True` ;)

PR #75, opened 2019-07-02
Post download link to logs, instead of uploading.

nlsdfnbch:
Should fix the issue of failing uploads. 

This requires an http server on the CI instance to run and serve the logs folder.

Such a service has been set up.


scenario_player/main.py
```diff
@@ -370,6 +370,7 @@ def construct_rc_message(log_fpath) -> str:
         message = f":x: Error while running scenario: {result}!"
         if exc:
             message += "\n```\n" + exc + "\n```"
+    message += f'\nLog can be downloaded from:\nscenario-player.ci.raiden.network/{str(log_fpath).split("/var/log/scenario-player/")[1]}'
```
ulope: I'd add `https://` to the URL.
Nitpick: Could be `log_path.relative_to(log_base_path)`

PR #92, opened 2019-07-11
Transaction Service Blueprint & Schema

nlsdfnbch:


scenario_player/services/transactions/app.py
```diff
@@ -0,0 +1,19 @@
...
+    transactions_blueprint, tokens_blueprint
+)
+
+from scenario_player.hooks.impl import HOOK_IMPL
+
+
+@HOOK_IMPL
+def register_blueprints():
+    """Register a list of blueprints with :mode:`raiden-scenario-player`."""
+    return [transactions_blueprint, tokens_blueprint]
```
nlsdfnbch: This hook implementation somehow is not called by the constructor on L17 below. I'm not sure where my spec/impl logic is faulty. Help is appreciated.

nlsdfnbch:
The Hook implementation is not honored, and I am not sure where the fault in my logic is. Since i've been staring at this for a bit now, a fresh pair of eyes may be of immense help.

scenario_player/services/transactions/app.py
karlb: Why should it be called? I don't see where the hooks are triggered.

nlsdfnbch: They're called here:
https://github.com/raiden-network/scenario-player/blob/7f5049e74a43a99393f61439125c9ff508bca952/scenario_player/hooks/__init__.py#L20

nlsdfnbch: Maybe I'm constructing the constant before the hook implementation can be registered?

nlsdfnbch:
After a dive into the source code, I believe to have found my mistake.

 It’s not enough to pass the module containing the `HookImplMarker` to `pluggy.PluginManager.register()`, since `plugy` doesn’t look for the *marker instance* itself, but for functions/methods *decorated with it inside the passed module*.

nlsdfnbch:
Rebased onto dev

nlsdfnbch:
tests passed locally. Investigating..

PR #100, opened 2019-07-16
Catch case where no IOU exists

czepluch:
The `assert_pfs_iou` task had a bug that allowed asserts on non existing IOUs to pass through the if statements and hence wasn't caught. 
This PR should ensure that we catch the cases where no IOU exists.

### DIFF
--- a/scenario_player/tasks/services.py
+++ b/scenario_player/tasks/services.py
@@ -289,8 +289,9 @@ class AssertPFSIoUTask(RESTAPIActionTask):
         if self._config.get("iou_exists", True) is False:
             if response_dict:
                 raise ScenarioAssertionError(f"Expected no IOU but got {response_dict}.")
-
-        if "amount" in self._config and "amount" in response_dict:
+        elif not response_dict:
+            raise ScenarioAssertionError(f"Expected IOU, but no IOU exists.")
+        else:
             exp_iou_amount = int(self._config["amount"])
             actual_iou_amount = int(response_dict["amount"])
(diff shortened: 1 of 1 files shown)


scenario_player/tasks/services.py
```diff
@@ -289,8 +289,9 @@ def _process_response(self, response_dict: dict):
         if self._config.get("iou_exists", True) is False:
             if response_dict:
                 raise ScenarioAssertionError(f"Expected no IOU but got {response_dict}.")
-
-        if "amount" in self._config and "amount" in response_dict:
+        elif not response_dict:
```
Dominik1999: the iou debug endpoint will always return a dict, see [here](https://github.com/raiden-network/raiden-services/blob/e254d894994f9db6d14e7fcc5fc5dbbf5d44af34/src/pathfinding_service/api.py#L390). Which means `response_dict` is always true, right?

Dominik1999: May use `elif "amount" in self._config and not "amount" in response_dict:` That way you cover the case where the PFS does not have an IOU and returns `{}` but the task expects an IOU

czepluch: > the iou debug endpoint will always return a dict, see [here](https://github.com/raiden-network/raiden-services/blob/e254d894994f9db6d14e7fcc5fc5dbbf5d44af34/src/pathfinding_service/api.py#L390). Which means `response_dict` is always true, right?

No, if you defint a dict `dct = {}` and then do `not dct` you get `True`, so this should be fine.

Issue #73, opened 2019-06-27
Drop support for Version 1 scenarios


nlsdfnbch:
Currently, version 1 scenarios are still supported - this, however, should be dropped in favor or version 2.

PR #105, opened 2019-07-19
Refactor .yaml config mapping

nlsdfnbch:
Finishes the initial refactoring plan started in scenario_player/scenario.py:

- Exposes scenario.yaml keys as attributes, for easier access
- Validates yaml sections on instantiating
- Implements Mapping interface
- Drops support for version 1 scenarios (#73)

This change is backwards compatible, as it does not touch existing code, but merely adds a new package.

### DIFF
--- a/scenario_player/utils/configuration/base.py
+++ b/scenario_player/utils/configuration/base.py
@@ -1,0 +1,41 @@
+from collections.abc import Mapping
+from typing import Optional, Union
+
+import structlog
+
+from scenario_player.exceptions.config import ConfigurationError
+
+log = structlog.get_logger(__name__)
+
+
+class ConfigMapping(Mapping):
+    def __init__(self, loaded_yaml: Mapping):
+        self.dict = loaded_yaml
+
+    def __getitem__(self, item):
+        return self.dict[item]
+
+    def __iter__(self):
+        return iter(self.dict)
+
+    def __len__(self):
+        return len(self.dict)
+
+    @staticmethod
+    def assert_option(expression, err: Optional[Union[str, Exception]] = None):
+        """Wrap `assert` to raise a ConfigurationError instead of an AssertionError."""
+        try:
+            assert expression
+        except AssertionError as e:
+            if err is None or isinstance(err, str):
+                raise ConfigurationError(err)
+            else:
+                exception = err
+            raise exception from e
+
+    def validate(self):
+        """Validate the configuration.
+
+        Assert that all required keys are present, and no mutually exclusive
+        options were set.
+        """
--- a/scenario_player/utils/configuration/base.py
+++ b/scenario_player/utils/configuration/base.py
@@ -9,6 +9,9 @@ log = structlog.get_logger(__name__)
 
 
 class ConfigMapping(Mapping):
+
+    CONFIGURATION_ERROR = ConfigurationError
+
     def __init__(self, loaded_yaml: Mapping):
         self.dict = loaded_yaml
 
@@ -21,14 +24,14 @@ class ConfigMapping(Mapping):
     def __len__(self):
         return len(self.dict)
 
-    @staticmethod
-    def assert_option(expression, err: Optional[Union[str, Exception]] = None):
+    @classmethod
+    def assert_option(cls, expression, err: Optional[Union[str, Exception]] = None):
         """Wrap `assert` to raise a ConfigurationError instead of an AssertionError."""
         try:
             assert expression
         except AssertionError as e:
             if err is None or isinstance(err, str):
-                raise ConfigurationError(err)
+                raise cls.CONFIGURATION_ERROR(err)
             else:
                 exception = err
             raise exception from e
--- a/scenario_player/utils/configuration/base.py
+++ b/scenario_player/utils/configuration/base.py
@@ -31,7 +31,7 @@ class ConfigMapping(Mapping):
             assert expression
         except AssertionError as e:
             if err is None or isinstance(err, str):
-                raise cls.CONFIGURATION_ERROR(err)
+                raise cls.CONFIGURATION_ERROR(err) from e
             else:
                 exception = err
             raise exception from e
--- a/scenario_player/utils/configuration/nodes.py
+++ b/scenario_player/utils/configuration/nodes.py
@@ -40,11 +40,6 @@ class NodesConfig(ConfigMapping):
         return self.get("node_options", {})
 
     @property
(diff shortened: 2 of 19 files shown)


nlsdfnbch:
Note that test files not located at `tests/unittest/utils/configuration/` have modifications to their syntax only, since I ran `black` on the tests folder. You can safely ignore these changes.

scenario_player/utils/configuration/base.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.dict[item]
+
+    def __iter__(self):
+        return iter(self.dict)
+
+    def __len__(self):
+        return len(self.dict)
+
+    @staticmethod
+    def assert_option(expression, err: Optional[Union[str, Exception]] = None):
```
ulope: This could be simplified if it would get the exception class to use from a class variable. Then you wouldn't need to override it on every inherited class.

scenario_player/utils/configuration/nodes.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.get("commands", {})
+
+    @staticmethod
+    def assert_option(expression, err: Optional[Union[str, Exception]] = None):
+        """Assert the given expression and raise a NodeConfigurationError if it fails."""
+        if isinstance(err, str):
+            err = NodeConfigurationError(err)
+        return ConfigMapping.assert_option(expression, err)
+
+    def validate(self):
```
ulope: This should also validate that `mode: managed` is specified (unless we want to make this optional since it's the only option for v2 currently).

scenario_player/utils/configuration/settings.py
```diff
@@ -0,0 +1,66 @@
...
+
+from scenario_player.constants import TIMEOUT
+from scenario_player.exceptions.config import ScenarioConfigurationError
+from scenario_player.utils import get_gas_price_strategy
+from scenario_player.utils.configuration.base import ConfigMapping
+
+log = structlog.get_logger(__name__)
+
+
+class SettingsConfig(ConfigMapping):
```
ulope: This is missing support for the `settings.services` section.

scenario_player/utils/configuration/token.py
```diff
@@ -0,0 +1,74 @@
...
+        mutual_exclusive_ops = ("address", "reuse")
+        if all(option in self.dict for option in mutual_exclusive_ops):
+            self.assert_option(
+                bool(self["address"]) != self["reuse"],
+                f"Token settings {mutual_exclusive_ops} are mutually exclusive.",
+            )
+
+    @property
+    def name(self):
+        if not self._name:
```
ulope: If reuse is true the name must be read from the reused contract.

scenario_player/utils/configuration/token.py
```diff
@@ -0,0 +1,74 @@
...
+                "name", f"Scenario Test Token {self._token_id!s} {now:%Y-%m-%dT%H:%M}"
+            )
+        return self._name
+
+    @property
+    def address(self):
+        return self.get("address")
+
+    @property
+    def block(self):
```
ulope: What does `block` do?

scenario_player/utils/configuration/token.py
```diff
@@ -0,0 +1,74 @@
...
+
+import structlog
+
+from scenario_player.exceptions.config import TokenConfigurationError
+from scenario_player.utils.configuration.base import ConfigMapping
+
+log = structlog.get_logger(__name__)
+
+
+class TokenConfig(ConfigMapping):
```
ulope: Missing options `balance_min` and `balance_fund`.

tests/unittests/utils/configuration/test_base.py
```diff
@@ -0,0 +1,34 @@
...
+    def test_assert_option_raises_exception_if_expression_is_false(self):
+        """:meth:`ConfigMapping.assert_option` raises an exception if expession is False."""
+
+    with pytest.raises(ConfigurationError):
+        ConfigMapping.assert_option(True is False)
+
+    def test_assert_option_completes_silently_if_expression_is_true(self):
+        """:meth:`ConfigMapping.assert_option` raises no exception if expession is True."""
+
+    ConfigMapping.assert_option(True is True)
```
ulope: a.k.a. `True` ;)

scenario_player/utils/configuration/nodes.py
```diff
@@ -0,0 +1,70 @@
...
+    """Thin wrapper around the 'nodes' setting section of a loaded scenario .yaml file.
+
+    Validates the givne config for missing values and mutually exclusive options.
+
+    :param loaded_yaml: The dict loaded from the scenario yaml file.
+    :param scenario_version: Version of the scenario yaml file.
+    """
+
+    def __init__(self, loaded_yaml: dict):
+        super(NodesConfig, self).__init__(loaded_yaml.get("nodes", {}))
```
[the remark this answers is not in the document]
nlsdfnbch: Yea, pycharm auto-completes that for me. :man_shrugging:

scenario_player/utils/configuration/nodes.py
nlsdfnbch: I've intentionally dropped it, since there were only two modes (external, managed) - since `external` was v1 only, it essentially boils down to being just `managed` - which is now implicit, and no longer required.

tests/unittests/utils/configuration/test_base.py
nlsdfnbch: long live the copy pasta!

scenario_player/utils/configuration/token.py
nlsdfnbch: I dont get the question? :thinking:

scenario_player/utils/configuration/token.py
nlsdfnbch: I would expect a `token_name` key to be present in the `token.info` file, that way I do not have to talk to the nodes at setup. 

This of course requires the step of saving token information to be extended accordingly.

scenario_player/utils/configuration/token.py
ulope: No I mean I don't know this setting. I don't think it exists ;)

nlsdfnbch: Aah..gotcha. Fixed!

scenario_player/utils/configuration/settings.py
nlsdfnbch: done.

PR #122, opened 2019-07-25
Update install/usage instructions

nlsdfnbch:
Updated README to include docker image pull/run instructions, and updated installation guide for `pip` users.


README.rst
```diff
@@ -3,24 +3,25 @@
...
-    # Clone the scenario player repo, and install on the Raiden dev version.
+    # Clone the scenario-player repository
     git clone http://github.com/raiden-network/scenario-player
-    cd scenario-player
-    pip install ".[dev]"
 
-Usage
------
+    # Install raiden's dev requirements.https://github.com/raiden-network/scenario-player/pull/122
+    pip install -r https://raw.githubusercontent.com/raiden-network/raiden/develop/requirements/requirements-dev.txt
```
Dominik1999: the requirements-dev.txt is not in the folder itself? Would that make sense? At least in the other Raiden Repos we have that always delivered within

README.rst
```diff
@@ -3,24 +3,25 @@
 
 Raiden Scenario Player
 ======================
-The Raiden Scneario Player is an integration testing tool written in Python 3. It allows testing of various test scenarios, and is
+The Raiden Scenario Player is an integration testing tool written in Python 3. It allows testing of various test scenarios, and is
 an integral component in the raiden test suite.
```
Dominik1999: is it "of the raiden test suite"? Can you spell Raiden with an capital "R"

README.rst
```diff
@@ -39,9 +40,28 @@ Using a ``pip`` installation::
...
+    docker pull raidennetwork/scenario-player:<tag>
+
+Where `<tag>` may be a specific version, git branch or `latest` for the last commit
+on `dev`, or `stable` for the last release on `master`.
+
+
+Usage
+=====
+
+Invoking the `scenario-platyr` from the cli can be done in one of the following
```
Dominik1999: there is a typo in `scenario-player`

README.rst
nlsdfnbch: The dev requirements arent our own, but raiden's - this way there is less hassle keeping the two in sync.

If `raiden` ever switches to `pyproject.toml` for requirements specifications we could set the scenario-player's `raiden` dependency to `raiden[dev]`, which would then install these requirements automatically, sans the extra step.

Until then I would keep it this way - otherwise it's guaranteed to be out of sync eventually.

README.rst
nlsdfnbch: I can and I shall! and `of` is correct, yes.

README.rst
nlsdfnbch: fixed!

PR #126, opened 2019-07-25
Refactor & Fix token network discovery check.

### LINKED ISSUE #125: Waiting for token network registration misses response check
When waiting for the token network to be discovered by the raiden nodes, there are two bugs which need to be taken care of:

The requested node is also the node we registered the token with - on fast connections, this returns a 200 OK, giving us the impression that the nodes already discovered the token network - which is false. **Token network discovery must be checked on a node different from the node used to register the token**.

Secondly, any response from the API causes the `while` loop to exit, since the API returns an APIError, which has a JSON payload (`{"errors": <error>}`). This payload is assigned to `self.token_network_address` and the SP never checks its value. This results in the faulty address being propagated downstream and blow up only after the tasks have already been started. **The loop waiting for token network discovery MUST check the returned value, and continue waiting if it's a `404`** - ensure this is *not* an infinite wait.

Source of all evil:
https://github.com/raiden-network/scenario-player/blob/12db4ec8f9037dd418ac10e8cb74780ff613008f/scenario_player/runner.py#L210:L222

Related REST API Endpoint:
https://github.com/raiden-network/raiden/blob/f034b2d001f31bfb00c334dfc4eed69f275153ea/raiden/api/rest.py#L780:L797

Logs supplied by @agatsoh (See line L670 for error message, and note how SP frolickingly continues on to its doom.) :
https://pastebin.com/kMf1Rwff

nlsdfnbch:
fixes #125

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -184,7 +184,7 @@ class ScenarioRunner:
             protocol=self.protocol, target_host=node, token_address=self.token_address
         )
 
-        started = time.time()
+        started = time.monotonic()
         elapsed = 0
         while elapsed < self.timeout:
             try:
@@ -201,7 +201,7 @@ class ScenarioRunner:
                 gevent.sleep(1)
 
                 # Update our elapsed time tracker.
-                elapsed = time.time() - started
+                elapsed = time.monotonic() - started
                 continue
 
             else:
@@ -214,7 +214,8 @@ class ScenarioRunner:
 
                 return data
 
-        # We could not assert that our token network was registered within an acceptable time frame.
+        # We could not assert that our token network was registered within an
+        # acceptable time frame.
         raise TokenNetworkDiscoveryTimeout
 
     def run_scenario(self):
@@ -256,7 +257,8 @@ class ScenarioRunner:
                 log.error("Couldn't register token with network", code=code, message=msg)
                 raise TokenRegistrationError(msg)
 
-        self.token_network_address = self.wait_for_token_network_discovery()
+        last_node = self.node_controller[-1].base_url
+        self.token_network_address = self.wait_for_token_network_discovery(last_node)
 
         log.info(
             "Received token network address", token_network_address=self.token_network_address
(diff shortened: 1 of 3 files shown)


scenario_player/runner.py
```diff
@@ -207,19 +256,7 @@ def run_scenario(self):
...
-        while self.token_network_address is None:
-            self.token_network_address = self.session.get(
-                API_URL_TOKEN_NETWORK_ADDRESS.format(
-                    protocol=self.protocol,
-                    target_host=first_node,
-                    token_address=self.token_address,
-                )
-            ).json()
-            gevent.sleep(1)
+        self.token_network_address = self.wait_for_token_network_discovery()
```
ulope: `node` parameter is missing.

tests/unittests/test_runner.py
```diff
@@ -0,0 +1,107 @@
+from unittest import mock
+
+import pytest
+import requests
+
+from scenario_player.runner import ScenarioRunner, TokenNetworkDiscoveryTimeout
+
+
+@pytest.fixture
+def response_class():
```
ulope: Nitpick: AFAICS this doesn't need to be a fixture

tests/unittests/test_runner.py
```diff
@@ -0,0 +1,107 @@
...
+        mock_request.return_value = response
+
+        # Somehow, wrapping a pytest.raises() block instead always raised a false positive.
+        try:
+            runner.wait_for_token_network_discovery("test_node")
+        except TypeError as e:
+            if not is_checksum:
+                # Should have raised the TypeError, all good.
+                return
+            raise AssertionError(f"DID RAISE {e!r}")
```
ulope: You could use `pytest.fail()` instead of raising the AssertionError

scenario_player/runner.py
nlsdfnbch: yikes.

tests/unittests/test_runner.py
nlsdfnbch: True, true.

PR #136, opened 2019-07-29
Add withdraw task

### LINKED ISSUE #72: Add task for `withdraw`
We need to add a task for `withdraw` in the scenario player. https://github.com/raiden-network/raiden/blob/develop/raiden/api/rest.py#L1054

ulope:
Fixes #72

### DIFF
--- a/scenario_player/tasks/channels.py
+++ b/scenario_player/tasks/channels.py
@@ -208,6 +208,7 @@ class AssertAllTask(ChannelActionTask):
                     f'Value mismatch for field "{field}". '
                     f"Not all values consumed, remaining: {channel_field_values}"
                 )
+        return response_dict
 
 
 class AssertSumTask(AssertAllTask):
@@ -237,3 +238,4 @@ class AssertSumTask(AssertAllTask):
                     f'Actual value: "{channel_value_sum}". '
                     f"Channels: {response_dict}"
                 )
+        return response_dict
--- a/scenario_player/tasks/channels.py
+++ b/scenario_player/tasks/channels.py
@@ -67,6 +67,22 @@ class DepositTask(ChannelActionTask):
         return dict(total_deposit=self._config["total_deposit"])
 
 
+class WithdrawTask(ChannelActionTask):
+    """ Perform a withdraw on the given channel.
+
+    Example usage::
+
+        # Withdraw ``100`` tokens from channel ``[Node 0] -> [Node 1]``
+        withdraw: {from: 0, to: 1, total_withdraw: 100}
+    """
+
+    _name = "withdraw"
+
+    @property
+    def _request_params(self):
+        return dict(total_withdraw=self._config["total_withdraw"])
+
+
 class TransferTask(ChannelActionTask):
     _name = "transfer"
     _url_template = "{protocol}://{target_host}/api/v1/payments/{token_address}/{partner_address}"
--- a/tests/unittests/tasks/test_channels.py
+++ b/tests/unittests/tasks/test_channels.py
@@ -1,0 +1,524 @@
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from scenario_player.exceptions import (
+    RESTAPIStatusMismatchError,
+    ScenarioAssertionError,
+    ScenarioError,
+)
+from scenario_player.tasks.channels import STORAGE_KEY_CHANNEL_INFO
+
+# TODO: Add tests for request timeouts
+
+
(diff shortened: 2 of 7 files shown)


ulope:
I think codecov is drunk...

tests/unittests/tasks/test_channels.py
```diff
@@ -0,0 +1,502 @@
...
+        "task_name",
+        "task_params",
+        "expected_exception",
+        "expected_exception_message",
+        "expected_req_method",
+        "expected_req_url",
+        "expected_req_body",
+        "resp_code",
+        "resp_json",
+    ),
```
nlsdfnbch: A general pointer for consistency: I state `argnames` as a string in tests, and pass `argvalues` as a keyword argument. I think it helps reading the parametrize mark.

tests/unittests/tasks/test_channels.py
```diff
@@ -0,0 +1,502 @@
...
+            'Expected all channels to be in "open" state.',
+            "GET",
+            f"http://0/api/v1/channels/0x1{1:039d}",
+            {},
+            200,
+            [{"state": "open"}, {"state": "closed"}],
+            id="assert_sum-state-mismatch",
+        ),
+    ],
+)
```
nlsdfnbch: Holy moly, that's a lot of parameters! Just an observation, carry on.

tests/unittests/tasks/test_channels.py
```diff
@@ -0,0 +1,502 @@
...
+        assert response == resp_json
+
+    if expected_req_method:
+        assert mocked_responses.calls[0].request.method == expected_req_method
+        assert mocked_responses.calls[0].request.url == expected_req_url
+        assert mocked_responses.calls[0].request.body.decode() == json.dumps(expected_req_body)
+        assert mocked_responses.calls[0].response.json() == resp_json
+
+
+def test_store_channel_info(mocked_responses, dummy_scenario_runner, api_task_by_name):
```
nlsdfnbch: This should be two tests. 

```
def test_store_channel_info_raises_exception_on_missing_key_config():
    ...

def test_store_channel_info_whatever_youre_testing_in_the_second_part():
     ...
```

tests/unittests/tasks/test_channels.py
```diff
@@ -0,0 +1,502 @@
...
+        "GET",
+        f"http://0/api/v1/channels/0x1{1:039d}/0x2{1:039d}",
+        json={"something": 1},
+        status=200,
+    )
+    task_instance = api_task_by_name("store_channel_info", {"from": 0, "to": 1, "key": "test"})
+
+    task_instance()
+
+    assert dummy_scenario_runner.task_storage[STORAGE_KEY_CHANNEL_INFO]["test"] == {"something": 1}
```
nlsdfnbch: From reading this, I'm not entirely sure what you're testing here - adding a comment or similar would help.

tests/unittests/tasks/test_channels.py
ulope: I'm not a fan of mangling the argnames into a single string. That's why we have tuples ;)

Also I'd argue we should be using the same code style here as everywhere else in raiden.

nlsdfnbch: > Also I'd argue we should be using the same code style here as everywhere else in raiden.

Yea, that's a pretty good argument.

scenario_player/tasks/channels.py
```diff
@@ -67,6 +67,14 @@ def _request_params(self):
         return dict(total_deposit=self._config["total_deposit"])
 
 
+class WithdrawTask(ChannelActionTask):
+    _name = "withdraw"
```
nlsdfnbch: Oh no, a task class without doc strings :<

tests/unittests/tasks/test_channels.py
```diff
@@ -0,0 +1,524 @@
...
+    because it raises an exception before getting to the request).
+
+    If ``expected_exception`` (and optional ``expected_exception_message``) is given we assert
+    this exception is actually raised.
+
+    Combinations of both of the above behaviours allow to assert on these three cases:
+      - Successful task execution
+      - Exception raised before a request is performed (e.g. violated precondition)
+      - Exception raised after a request is performed (e.g. processing return value)
+    """
```
nlsdfnbch: mmmh, delicious docs. :+1:

czepluch:
Would be nice if you could also add an entry in the example file.

Issue #62, opened 2019-06-24
RPC Service - Running services concurrently


nlsdfnbch:
Analysis
=======
Current state
----------------
The scenario player runs scenarios one at a time, increasing the overall run-time with each added scenario. Running them concurrently is not an option, due to `nonce` errors.

Problem
-----------
Running scenarios concurrently is not possible with the current SP version is not possible, due to errors in `nonce` generation: `raiden` expects complete and sole control over an account. Since several scenarios start several `raiden` instances, this assumption does not hold, and duplicate nonces are generated, causing our scenarios to error.

Solution
----------
Replace the use of `JSONRPCClient` in the `ScenarioRunner` with a micro service which manages RPC client instances for all currently running scenarios.

It must, essentially, take over all duties that include a `JSONRPCClient.send_transaction()` call.

Service Definition
==============

Domain
----------
The new service owns the following domain:

- Sending a transaction
- Deploy a token contract
- Mint tokens for a token contract 
- Update the allowance for a User Deposit Contract

Dependencies
------------------
Requires the following packages, in order to run correctly:

- `raiden`
- `raiden_contracts`
- `web3`
- `ethutils`

Requires no additional `SPaaS` services to run.

Interface
-----------

The following endpoints are exposed:

- `/rpc/client`
  - `POST` - creates a new `JSONRPCClient` instance, with a given configuration.
  - `DELETE` - removes an existing rpc instance from the client registry.
- `/rpc/contract`
  - `POST` - deploys a new token contract.
- `/rpc/contract/mint`
  - `POST` mints new tokens for a given address.
- `/rpc/tcontract/allowance`
  - `POST`  - updates the allowance for the given UDC contract.

Tasks
=======
- [ ] Testing

palango:
Closing. See https://github.com/raiden-network/scenario-player/issues/495

PR #201, opened 2019-08-09
RPC Service - Allow concurrent Scenario Runs

### LINKED ISSUE #208: Daemonizer

nlsdfnbch:
Finishes #62 and closes #208 

~Allows running multiple scenarios concurrently, when passing the `--spaas-no-teardown` flag to the SP cli command.~

This no longer run dameonized processes, but instead creates `systemd` user-land services.

Before running the SP, installation is required:

```
spaas install
```

Which creates service files and enables them for the user.

Removing them is simple as well, and takes care of stopping and disabling them first:
```
spaas remove
```


scenario_player/services/utils/factories.py
```diff
@@ -45,3 +51,118 @@ def construct_flask_app(
...
+    try:
+        if os.fork() > 0:
+            raise SystemExit(0)
+    except OSError:
+        raise RuntimeError("fork #2 failed.")
+
+    # Flush I/O buffers
+    sys.stdout.flush()
+    sys.stderr.flush()
+
```
schmir-at-bb: you should close all file descriptors. see https://stackoverflow.com/questions/17954432/creating-a-daemon-in-linux/17955149#17955149
there's no need for the dup2 calls then

nlsdfnbch:
> I would probably look into using a ready-made library, maybe https://github.com/thesharp/daemonize

Done. Thanks for the review!

Issue #218, opened 2019-08-12
The default minted amount of RDN Tokens of the UDC is not sufficient for some scenarios


Dominik1999:
At the moment nodes seem to get 1000 Test-RDN Rai tokens at the start of a scenario when `udc` is enabled. Nodes for the basic functionalities scenario, see [here](https://github.com/raiden-network/raiden/pull/4499), needs more of these tokens. 

Can this be a quick fix? Then we can test the new basic functionality scenario.

nlsdfnbch:
The template you used is for Service Epics, which this clearly is not - I did add some bug report templates now, in order to avoid this sort of confusion. I went ahead and updated your issue.

>Can this be a quick fix?

To clarify: There are no "quick fixes" - there are fixes, and there are features. We can release at any commit, so there should be no need to call these fixes "quick fixes", as that implies they're somehow dirty hacks. And we don't do dirty hacks in this hood.

Back to the topic at hand:

I believe this issue encompasses two actual issues, a fix and a feature request:

- [ ] FIX: Update the (currently) fixed amount of minted RDN tokens, so that the new scenarios can run.
- [ ] FEATURE: Implement the option to specify the amount of RDN tokens minted in the scenario yaml.


As the work required to implement the feature is arguably small, I would suggest skipping the fix and going for the feature directly.

christianbrb:
@Dominik1999 is testing with #135

christianbrb:
@Dominik1999 
> yes I tested it locally yesterday and it worked

@nlsdfnbch This can be released

nlsdfnbch:
Release with `0.3.0` (https://github.com/raiden-network/scenario-player/pull/232)

PR #221, opened 2019-08-13
Implement UDC balance_per_node and max_funding yaml options.

nlsdfnbch:
Implements feature as requested per #218 :

2 new keys are available in the `udc` section of the yaml file:

```yaml
settings:
  services:
    udc:
      deposit: true
      # The number of RDN/UDC Tokens required by each node
      balance_per_node: 5000
      # The maximum funding amount to deposit - defaults to `balance_per_node` if not given.
      max_funding: 5000
```

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -327,10 +327,10 @@ class ScenarioRunner:
         self.node_controller.initialize_nodes()
         node_addresses = self.node_controller.addresses
         node_count = len(self.node_controller)
-        node_balances = {address: self.client.balance(address) for address in node_addresses}
+        balance_per_nodes = {address: self.client.balance(address) for address in node_addresses}
         low_balances = {
             address: balance
-            for address, balance in node_balances.items()
+            for address, balance in balance_per_nodes.items()
             if balance < NODE_ACCOUNT_BALANCE_MIN
         }
         if low_balances:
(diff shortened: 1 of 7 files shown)


scenario_player/runner.py
```diff
@@ -327,10 +327,10 @@ def _initialize_nodes(
         self.node_controller.initialize_nodes()
         node_addresses = self.node_controller.addresses
         node_count = len(self.node_controller)
-        node_balances = {address: self.client.balance(address) for address in node_addresses}
+        balance_per_nodes = {address: self.client.balance(address) for address in node_addresses}
```
czepluch: Shouldn't it be `node` instead of `nodes`? At least that's what you used everywhere else and it's correct English.

scenario_player/runner.py
```diff
@@ -327,10 +327,10 @@ def _initialize_nodes(
...
         low_balances = {
             address: balance
-            for address, balance in node_balances.items()
+            for address, balance in balance_per_nodes.items()
```
czepluch: same as above

scenario_player/runner.py
nlsdfnbch: Yupp, you're right - missed the `s` when replacing (`node_balance` instead of `node_balances`) so that was never removed. fixed!

christianbrb:
@nlsdfnbch Is this backward compatible -> does the old functionality still works or do we need to adjust the scenarios?

> At the moment nodes seem to get 1000 Test-RDN Rai tokens at the start of a scenario when udc is enabled.

Issue #187, opened 2019-08-05
Properly handle a bad password during account unlock


hackaugusto:
I opened issue https://github.com/raiden-network/scenario-player/issues/185 because I assumed the unlock was failing for some reason, turns out that the password is given in the command line instead of a file. The unlock simply failed because the password was incorrect.

Handling the exception and printing a error message would be nice.

Issue #192, opened 2019-08-06
Improve PW handling


nlsdfnbch:
Passwords for accounts need pimping:

- Allow passing password files instead of plain text strings
- Allow piping passwords from stdin
- Properly handle bad passwords during account unlock
- Allow passing empty passwords

nlsdfnbch:
Merged to `dev`

PR #222, opened 2019-08-13
Improve pw handling

eorituz:
@nlsdfnbch pls check if you're happy with the test structure, dunno if we have a "style guide" for them.


eorituz:
This should cover everything in Issue #192

tests/unittests/cli/test_cli.py
```diff
@@ -0,0 +1,102 @@
...
+from scenario_player.exceptions.cli import WrongPassword
+
+
+@pytest.fixture(scope="module")
+def runner():
+    return CliRunner()
+
+
+KEYSTORE_PATH = str(Path(__file__).resolve().parents[0].joinpath("keystore"))
+CLI_ARGS = [
```
nlsdfnbch:
```suggestion
CLI_ARGS = f"--chain http://geth.goerli.ethnodes.brainbot.com:8545 run --keystore-file {KEYSTORE_PATH.joinpath('UTC--1')} --no-ui  {{pw_option}} {Path(__file__).resolve().parent.joinpath('scenario', 'join-network-scenario-J1.yaml')}"
```

And in tests use `str.format()` instead of list concatenation:

```python
...
result = runner.invoke(main.main, CLI_ARGS.format(pw_option="--password 123").split(" ")
...
```

eorituz:
This PR should also fix #187

scenario_player/main.py
```diff
@@ -89,6 +91,22 @@ def load_account_obj(keystore_file, password):
...
 
+def get_password(password, password_file):
+    if password_file:
+        password = open(password_file, "r").read().strip()
+    if password == password_file is None:
+        password = click.prompt(text="Please enter your password: ", hide_input=True)
+    return password
+
+
+def get_account(keystore_file, password):
```
nlsdfnbch: Looking at this I just realized this could be moved into `load_account_obj` instead of adding a func that, essentially, only adds a try/except block.

```
def load_account_obj(keystore_file, password):
    with open(keystore_file, "r") as keystore:
        try:
            account = Account(json.load(keystore), password, keystore_file)
        except ValueError:
            raise WrongPassword
        log.info("Using account", account=to_checksum_address(account.address))
        return account
```

PR #237, opened 2019-08-15
Added description to example yaml

### LINKED ISSUE #227: Fee setting in SP
Analysis
=======
Current State
-----------------
Raiden nodes now have mediation fee settings:
```
Mediation Fee Options:
  --flat-fee FEEAMOUNT            Flat fee required for every mediation in wei of the mediated token.  [default: 0]
  --proportional-fee INTEGER      Mediation fee as ratio of mediated amount in parts-per-million (10^-6).  [default:
                                  0]
  --max-imbalance-fee FEEAMOUNT   Set the worst-case imbalance fee in wei of the mediated token. (Preview)  [default:
                                  0]
```

Problem
-----------
The mediation fees are not settable in the scenario player. We need the mediation fees for the following scenario https://github.com/raiden-network/raiden/issues/4594

Solution
----------
Add the settings to the known settings options.

----------------------------------------------------------------------
User Stories
==========
*Describe the requested feature in user stories, describing one behavior at a time, with one story each. These are later used to test and assert the correctness of the feature, so be detailed and thorough when writing these. They may be updated later, during the course of the discussion of this issue.*

eorituz:
This PR is basically the "documentation" for #227 and therefore fixes #227

### DIFF
--- a/examples/scenario-example-v2.yaml
+++ b/examples/scenario-example-v2.yaml
@@ -60,8 +60,10 @@ nodes:
   ## Options to apply to all nodes
   ## Option names correspond to Raiden cli options without the leading double dash (`--`)
   default_options:
+    ## Sets Raiden config parameters.
     gas-price: fast
-    # environment-type: development
+    #environment-type: development
+    #flat-fee: 10
     #registry-contract-address: "0xbfa863Ac58a3E0A82B58a8e958F2752Bfb573388"
     #discovery-contract-address: "0xb2646EB8270a4de7451f6F7b259FdE17DBCeedc9"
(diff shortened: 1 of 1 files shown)


christianbrb:
Thanks :-)

Should we also document the two other options even though just one of them can be used at a time?
```
--proportional-fee INTEGER
--max-imbalance-fee FEEAMOUNT

```

eorituz:
I don't think we need to add all individual Raiden config parameters. 
That's why I added the sentence `## Sets Raiden config parameters.`. You can get all Raiden config parameters by running `raiden --help`.

PR #239, opened 2019-08-15
Waiting for all nodes to discover token network

### LINKED ISSUE #132: Token Network Discovery: Check all nodes, instead of one
As pointed out by @hackaugusto [here](https://github.com/raiden-network/scenario-player/issues/125#issuecomment-515426458), checking only a single node for token network discovery is insufficient, due to various race conditions.

Hence, check all nodes instead of the last created one.

Related code:
https://github.com/raiden-network/scenario-player/blob/76194049e8e2ca46d18c627fb95090b2919456a7/scenario_player/runner.py#L173:L219

eorituz:
This should resolve #132
@nlsdfnbch do you think the tests are still sufficient or do we need more?

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -157,8 +157,8 @@ class ScenarioRunner:
                 f'The scenario requested chain "{chain_name}" for which no RPC-URL is known.'
             )
 
-    def wait_for_token_network_discovery(self, node):
-        """Check for token network discovery with the given `node`.
+    def wait_for_token_network_discovery(self, node_urls):
+        """Check for token network discovery for all `nodes`.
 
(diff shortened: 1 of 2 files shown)


nlsdfnbch:
@eorituz no, they're not - we need an assertion/test that all nodes are queried, not just one (since that's what this fix is about).

eorituz:
Mhh as far as I got it this test querries all nodes:

```
    def test_wait_for_token_network_discovery_returns_checksum_when_network_is_discovered(
        self, mock_request, response, runner
    ):
        mock_request.return_value = response
        actual = runner.wait_for_token_network_discovery(["test_node1", "test_node2"])
        assert actual == response.payload

Issue #238, opened 2019-08-15
Transaction [txhash] failed


czepluch:
Description
=========
Today when trying to start a new scenario I run into the error shown in the attached logfile.
[scenario-player-run_pfs7_multiple_payments_2019-08-15T16:54:56.log](https://github.com/raiden-network/scenario-player/files/3505967/scenario-player-run_pfs7_multiple_payments_2019-08-15T16.54.56.log)

It seems that a tx fails when trying to call `deposit` in the UDC contract. The problem also happens on old scenarios that I've run earlier today.

I tried installing the SP in a fresh venv too, but the problems persists.

This is on latest dev branch of the SP and of Raiden.

Expected Behavior
===============
I expected the SP to start the scenario as usual and to fund the nodes correctly.

czepluch:
Tried to use both the bb eth node vpn and my dappcon node, but same problem occurs.

eorituz:
So I can't reproduce the error.
Your error code points to a reverted [onchain Tx](https://goerli.etherscan.io/tx/0xb508c34739260b6031e96deed105783edc0306c9b7addbe470789656c2f91ce4)
No clue what goes wrong here, but it works fine for me atm.

Dominik1999:
I reproduced the error. I first successfully ran a scenario, then, when I simply want to re-run it, I get the same error. 

```
{"result": "scenario error", "event": "Run finished", "logger": "scenario_player.main", "level": "error", "timestamp": "2019-08-16 09:46:51.021767", "exception": "Traceback (most recent call last):\n  File \"/home/pirate/PycharmProjects/ScenarioPlayer/scenario-player/scenario_player/main.py\", line 179, in run\n    runner.run_scenario()\n  File \"/home/pirate/PycharmProjects/ScenarioPlayer/scenario-player/scenario_player/runner.py\", line 224, in run_scenario\n    wait_for_txs(self.client, fund_tx | ud_token_tx | mint_tx)\n  File \"/home/pirate/PycharmProjects/ScenarioPlayer/scenario-player/scenario_player/utils/legacy.py\", line 193, in wait_for_txs\n    raise ScenarioTxError(f\"Transaction {encode_hex(txhash)} failed.\")\nscenario_player.exceptions.legacy.ScenarioTxError: Transaction 0xe8195f306388726e7810eb84d5e1e1bac96fc884eb67ad13fc3c09998bbd77e3 failed."}
```

Now it constantly fails with this error

[...]

PR #249, opened 2019-08-19
Fixed calculation of deposit_amount

eorituz:
Since jacobs scenario is the first one to [use](https://github.com/raiden-network/raiden/pull/4619/files#diff-8f0eba5239113fc61a0d292756667b6bR13) `balance_per_node` 
We never tested the calculation for depositing more tokens.

Since the UDC contract requires a deposit greater than the total_deposit 
``` 
function deposit(address beneficiary, uint256 new_total_deposit)
        external
    {
        require(new_total_deposit > total_deposit[beneficiary], "deposit not increasing");

        // Calculate the actual amount of tokens that will be transferred
        uint256 added_deposit = new_total_deposit - total_deposit[beneficiary];

        balances[beneficiary] += added_deposit;
        total_deposit[beneficiary] += added_deposit;

        // Update whole_balance, but take care against overflows.
        require(whole_balance + added_deposit >= whole_balance, "overflowing deposit");
        whole_balance += added_deposit;

        // Decline deposit if the whole balance is bigger than the limit.
        require(whole_balance <= whole_balance_limit, "too much deposit");

        // Actual transfer.
        require(token.transferFrom(msg.sender, address(this), added_deposit), "tokens didn't transfer");
```

I needed to change the calculation of the amount that needs to be deposited

This is a fix for Issue #238


scenario_player/utils/token.py
```diff
@@ -409,6 +414,6 @@ def deposit(self, target_address) -> Union[str, None]:
             return
 
         log.debug("deposit call required - insufficient funds")
-        deposit_amount = max_funding - balance
+        deposit_amount = total_deposit + max(max_funding, min_deposit) - balance
```
nlsdfnbch: `max()` should not be necessary here - there should be a check<sup>[suggestion](#myfootnote1)</sup>

 in [this class](https://github.com/raiden-network/scenario-player/blob/master/scenario_player/utils/configuration/settings.py#L40:L61) instead, that ensures that `max_funding >= balance_per_node` when loading the scenario yaml.

Otherwise, this would silently use a value that may not be expected by the user.

-----------------------------------------------------------------
<a name="myfootnote1">Suggested validation:</a>:

```
def validate(self):
    self.assert_option(self.max_funding < self.balance_per_node, "max_funding must be >= balance_per_node value!")
```

I trust you can write a test for that just fine. ;P

nlsdfnbch:
That would've taken me days to gifure out - good catch! Just a small change required, then we're good to merge this.

tests/unittests/utils/configuration/test_settings.py
```diff
@@ -158,3 +157,14 @@ def test_attributes_whose_key_is_absent_return_expected_default(
...
+    def test_balance_per_node_bigger_than_max_funding(self, minimal_yaml_dict):
+        minimal_yaml_dict["max_funding"] = 5000
+        minimal_yaml_dict["balance_per_node"] = 5001
+        assert UDCTokenSettings(minimal_yaml_dict).CONFIGURATION_ERROR == ConfigurationError
+
+    # FIXME this test doesn't work for whatever reason
+    def test_insufficient_minting(self, minimal_yaml_dict):
+        minimal_yaml_dict["max_funding"] = 5000
+        minimal_yaml_dict["token"]["min_balance"] = 4999
+        assert UDCTokenSettings(minimal_yaml_dict).CONFIGURATION_ERROR == ConfigurationError
```
nlsdfnbch: I think there's a misconception about that `.CONFIGURATION_ERROR` is supposed to be:
That's the exception type to be invoked by `.validate` - it does *not* store exceptions raised during `validate()`. 

Instead, this should be:

```suggestion
with pytest.raises(ConfigurationError):
    UDCTokenSettings(minimal_yaml_dict)
```

It would also be reasonable to create a `UDCTokenConfigError`, and assign that to the class's `CONFIGURATION_ERROR` attribute, and change `pytest.raises` to look for that instead.

scenario_player/utils/configuration/settings.py
```diff
@@ -41,8 +41,30 @@ class UDCTokenSettings(ConfigMapping):
...
+        try:
+            min_balance = loaded_yaml["token"]["min_balance"]
+        except (KeyError, TypeError):
+            min_balance = DEFAULT_TOKEN_BALANCE_FUND
+        # The amount of Tokens minted need to be greater than the amount of tokens
+        # that will be deposited into the UDC
+        self.assert_option(
+            self.max_funding <= min_balance,
+            "Token.max_funding needs to be greater than the Deposit amount."
+            f"The Deposit amount is max({DEFAULT_TOKEN_BALANCE_FUND}, "
```
nlsdfnbch: This needs to be the actual value of `token.max_funding` - as as you've already pointed out, this check would be easier and cleaner to do on the `ScenarioYAML` class.

tests/unittests/utils/configuration/test_settings.py
eorituz: Yep, it was a misconception. Changed it.

scenario_player/scenario.py
```diff
@@ -33,10 +34,19 @@ def __init__(self, yaml_path: pathlib.Path, data_path: pathlib.Path) -> None:
...
 
         self.gas_limit = GAS_LIMIT_FOR_TOKEN_CONTRACT_CALL * 2
 
     @property
     def name(self) -> str:
         """Return the name of the scenario file, sans extension."""
         return self.path.stem
+
+    def validate(self):
+        # The amount of Tokens minted need to be greater than the amount of tokens
```
nlsdfnbch: That can go into a the method doc string, since it would be interesting to know for anyone looking up this method (once we render them somewhere).

scenario_player/utils/configuration/base.py
```diff
@@ -49,7 +49,7 @@ def assert_option(cls, expression, err: Optional[Union[str, Exception]] = None):
                 exception = err
             raise exception from e
 
-    def validate(self):
+    def validate(self, **args):
```
nlsdfnbch: This should no longer be required, no?

scenario_player/utils/configuration/settings.py
```diff
@@ -41,8 +45,12 @@ class UDCTokenSettings(ConfigMapping):
     def __init__(self, loaded_yaml: dict):
         udc_settings = ((loaded_yaml.get("settings") or {}).get("services") or {}).get("udc") or {}
         super(UDCTokenSettings, self).__init__(udc_settings.get("token"))
+        self.validate(loaded_yaml)
         print(self.dict)
 
+    def validate(self, loaded_yaml):
```
nlsdfnbch: the `loaded_yaml` should no longer be required, no?

tests/unittests/utils/configuration/test_settings.py
```diff
@@ -158,3 +169,16 @@ def test_attributes_whose_key_is_absent_return_expected_default(
         config = UDCTokenSettings(minimal_yaml_dict)
         MISSING = object()
         assert getattr(config, key, MISSING) == expected
+
+    def test_balance_per_node_bigger_than_max_funding(self, minimal_yaml_dict):
```
nlsdfnbch:
```suggestion
    def test_balance_per_node_must_not_be_greater_than_max_funding(self, minimal_yaml_dict):
```

Otherwise, it sounds like your assertin that balance_per_node should be greater than token max_funding.

scenario_player/utils/configuration/settings.py
```diff
@@ -41,8 +45,12 @@ class UDCTokenSettings(ConfigMapping):
     def __init__(self, loaded_yaml: dict):
         udc_settings = ((loaded_yaml.get("settings") or {}).get("services") or {}).get("udc") or {}
         super(UDCTokenSettings, self).__init__(udc_settings.get("token"))
+        self.validate(loaded_yaml)
         print(self.dict)
```
nlsdfnbch: ew, del that `print`, however that got in there.

PR #267, opened 2019-08-23
Fix Pack-logs command.

### LINKED ISSUE #269: Separate logging to one file per scenario
Analysis
=======
Current State
-----------------
When the scenario player starts a run, it writes all logs to one file rather than create separate files. This used to be different.

Problem
-----------
It is difficult to get a good overview of what went wrong in a scenario when all scenarios are mixed together in the same log file at the moment. 

Solution
----------
The above problem would be solved if each scenario has it's own log file per run as it used to be in the past.

nlsdfnbch:
Fixed #269

Previously tried to load run number from a non-existing file, which made it default to `0` (i.e. "get all logs"), causing all node folders to be packed. Fixed that.

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -333,9 +333,10 @@ def reclaim_eth(ctx, min_age, password, password_file, keystore_file):
 @click.argument("scenario-file", type=click.Path(exists=True, dir_okay=False), required=True)
 @click.pass_context
 def pack_logs(ctx, scenario_file, post_to_rocket, pack_n_latest, target_dir):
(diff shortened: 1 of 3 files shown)


christianbrb:
Is this PR fixing #267 ?

agatsoh:
Just ran multiple runs of the scenario and it accurately downloads the nightlies and makes a log file per scenario run, but fails with [this](https://gist.github.com/agatsoh/8e80fc7e2fd237c4f4952b6dfb57c480) error.

PR #275, opened 2019-09-02
Fix token balance check and UDC allowance minting

ulope:
Fixes: #238

### DIFF
--- a/scenario_player/utils/token.py
+++ b/scenario_player/utils/token.py
@@ -61,7 +61,9 @@ class Contract:
         log.info(f"'{action}' call succeeded", tx_hash=tx_hash)
         return decode_hex(tx_hash)
 
-    def mint(self, target_address, **kwargs) -> Union[str, None]:
+    def mint(
+        self, target_address, required_balance=None, max_fund_amount=None, **kwargs
+    ) -> Union[str, None]:
         """Mint new tokens for the given `target_address`.
 
         The amount of tokens depends on the scenario yaml's settings, and defaults to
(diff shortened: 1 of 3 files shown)


scenario_player/utils/token.py
```diff
@@ -349,6 +358,11 @@ def allowance(self):
             self._local_rpc_client.address, self.address
         ).call()
 
+    @property
+    def balance(self):
+        """Proxy the balance call to the UDTC."""
+        return self.token_proxy.contract.functions.balanceOf(self.ud_token_address).call()
```
hackaugusto: I'm not 100% convinced on this fix.

Here is a test script. I just wanted to make sure that a subclass could use `super()` on a super class, and the code of the super class would still use the subclass properties:

```python
class A:
    @property
    def balance(self):
        return 1

    def mint(self):
        return self.balance

class B(A):
    def mint(self):
        return super().mint()

    @property
    def balance(self):
        return 2

a, b = A(), B()
print(a.balance, a.mint(), b.balance, b.mint())
```
The above will print `1 1 2 2`, meaning that it does. 

This happens on our code [here](https://github.com/raiden-network/scenario-player/blob/0701bb986f47e1ec4a4fb7a469157826da1993e2/scenario_player/runner.py#L316), the call `self.udc.mint(our_address)` will use this balance, which is the balance of `ud_token_address`. If I understood this correctly, the formula `self.config.token.max_funding - balance` will eventually only negative values.

ulope: > super class would still use the subclass properties

If that wasn't the case OO would be pretty broken in Python ;)

With the additional changes it should now be correct (since the balance of the SP account is used to `deposit` into the separate Raiden nodes).

ulope:
Codecov seems to be stuck. Since the tests are green and it's approved I'll force merge this.

Dominik1999:
Yeah!

PR #278, opened 2019-09-04
Fix UDTC allowance calculation

ulope:
The UDTC allowance was calculated as a delta of required - current balance, but the `approve` call expects an absolute value. 
This lead to insufficient balance on the last `deposit` call.

Fixes: #238

### DIFF
--- a/scenario_player/utils/token.py
+++ b/scenario_player/utils/token.py
@@ -27,6 +27,13 @@ class Contract:
         self.interface = ServiceInterface(runner.yaml.spaas)
         self.gas_limit = GAS_LIMIT_FOR_TOKEN_CONTRACT_CALL * 2
 
+    def __repr__(self):
+        return f"<{self.name}>"
+
+    @property
+    def name(self):
+        return f"{self.__class__.__name__}@{to_checksum_address(self.address)}"
+
     @property
     def client_id(self):
         return self.config.spaas.rpc.client_id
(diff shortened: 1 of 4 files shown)


scenario_player/utils/token.py
```diff
@@ -399,19 +405,20 @@ def update_allowance(self) -> Union[Tuple[str, int], None]:
         required_allowance = self.config.settings.services.udc.token.balance_per_node * node_count
 
         log.debug(
```
palango: Should you use `local_log` here as well?

ulope: Probably yeah, I'll make an issue about it.

Dominik1999:
Yeah!

scenario_player/utils/token.py
```diff
@@ -399,19 +405,20 @@ def update_allowance(self) -> Union[Tuple[str, int], None]:
...
-            log.debug("allowance update call not required - sufficient allowance")
+            log.debug("UDTC allowance sufficient")
             return
 
-        log.debug("allowance update call required - insufficient allowance")
-        allow_amount = required_allowance - udt_allowance
+        log.debug("UDTC allowance insufficient, updating")
         params = {
-            "amount": allow_amount,
+            "amount": required_allowance,
```
hackaugusto: isn't this a regression?

ulope: No, that's the bugfix

hackaugusto: Sorry, let me rephrase the question. What was wrong in the previous version?

PR #306, opened 2019-09-10
Added reveal_timeout to channel open task

hackaugusto:
This should wait, atm this is not available in the raiden rest api


nlsdfnbch:
Covered by https://github.com/raiden-network/scenario-player/pull/358

Issue #277, opened 2019-09-04
Raiden node private keys are not unique


ulope:
Description
=========
The Raiden node private keys are generated from the following seed:
```
f"{self._runner.yaml.name}-{self._runner.run_number}-{self._index}"
```

This means that the private key is (in effect) only dependant on the scenario name. This leads to the same private keys being generated for the same scenarios on different machines.

Expected Behavior
===============
The Node private keys need to be deterministic but unique.

Proposed solution
=============

Store a per-instance random seed in the scenario player data dir and incorporate that into the private key seed. This has the advantage over alternatives like hostname or hardware addresses (e.g. MAC address) that it's easily resettable and also not influenced by running the SP inside docker etc.

PR #316, opened 2019-09-11
[FIX-#277] Make node keys unique

ulope:
Ensures node keys are unique between multiple users of the scenario
player by using a per-installation (rather per data-dir) seed value.

Fixes: #277


ulope:
Test added

PR #337, opened 2019-09-19
Allow passing multiple values to Raiden CLI Option as list

### LINKED ISSUE #336: Two arguments given to an option passes wrong address
Description
=========
For https://github.com/raiden-network/raiden/issues/4890 we need to pass two arguments to `flat-fee`. Doing so, however, results in a `Error: Invalid value for "--flat-fee": Address must be EIP55 checksummed` during start up of the nodes used in a scenario. According to the `.stdout` file of the node, the arguments are parsed correctly to the `raiden` command. 

Above is with ` flat-fee: 0x62083c80353Df771426D209eF578619EE68D5C7A 100`

I also tried with `flat-fee: "0x62083c80353Df771426D209eF578619EE68D5C7A" 100`, but that doesn't even start the SP.

Expected Behavior
===============
Expected behaviour is that multiple arguments can be provided to an option and interpreted correctly.

nlsdfnbch:
Some options now require multiple values. This fix takes care of checking if the supplied option's value is a list, and unpacks it before extending the node command.

To make use of this, formulate the option as follows:
```yaml
---
my-amazing-cli-flag:
  - value1
  - value2
```

Fixes #336

### DIFF
--- a/scenario_player/node_support.py
+++ b/scenario_player/node_support.py
@@ -323,16 +323,25 @@ class NodeRunner:
                 # already handled above
                 continue
             if option_name in self._options:
-                cmd.extend([f"--{option_name}", self._options[option_name]])
+                option_value = self._options[option_name]
+                if isinstance(option_value, list):
+                    cmd.extend([f"--{option_name}", *self._options[option_name]])
+                else:
+                    cmd.extend([f"--{option_name}", self._options[option_name]])
 
         remaining_option_candidates = (
             KNOWN_OPTIONS - MANAGED_CONFIG_OPTIONS - MANAGED_CONFIG_OPTIONS_OVERRIDABLE
         )
         for option_name in remaining_option_candidates:
             if option_name in self._options:
-                cmd.append(f"--{option_name}")
                 if option_name not in FLAG_OPTIONS:
-                    cmd.append(self._options[option_name])
+                    option_value = self._options[option_name]
+                    if isinstance(option_value, list):
+                        cmd.extend([f"--{option_name}", *self._options[option_name]])
+                    else:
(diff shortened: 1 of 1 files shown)


czepluch:
Checked out the branch locally and raiden was attempted to be started like this: ` --flat-fee [559664771866394222468386075950557960554851163258, 100]` for the flat fee.

So it looks like the arguments are parsed wrongly and that the address is interpreted wrongly too.

ulope:
@czepluch did you put quotes around the address? If not the tank parser unpacks it into an int. (That’s what it looks like in your case)

nlsdfnbch:
> Checked out the branch locally and raiden was attempted to be started like this: ` --flat-fee [559664771866394222468386075950557960554851163258, 100]` for the flat fee.
> 
> So it looks like the arguments are parsed wrongly and that the address is interpreted wrongly too.

fixed that - I initially forgot to apply the fix to a second code block. Should now work as expected.

czepluch:
Still doesn't work for me. I added " " around the address and ran it again with the new changes. Getting `Error: Invalid value for "--flat-fee": Address must be 0x prefixed` now even though the logs show Raiden so start with the correct arguments.

scenario_player/node_support.py
```diff
@@ -332,7 +336,11 @@ def _command(self):
             if option_name in self._options:
                 cmd.append(f"--{option_name}")
                 if option_name not in FLAG_OPTIONS:
-                    cmd.append(self._options[option_name])
+                    option_value = self._options[option_name]
+                    if isinstance(option_value, list):
```
ulope: This doesn't look right. The `--{option_name}` already get's appended in line 337.

czepluch:
Seems to work now.

Issue #333, opened 2019-09-16
All scenarios asserting on amount of ChannelClosed events fail


czepluch:
Description
=========
It seems that after we started using the same token for all scenario runs that the scenarios that use `AssertBlockchainEventsTask` (All MS and BF) start failing if they assert on the number of `ChannelClosed` events. This is most likely due to the fact that using the same token network generates more events and hence the number of events grow over time for the same token network.

Expected Behavior
===============
I would expect that there is a way to assert only on the amount of events for a specific channel. Alternatively a solution for now could be to not assert on the expected amount of `num_events`. But this could also mean that there are other bugs we won't catch in the future.

PR #340, opened 2019-09-20
[FEAT-#333] Allow assert_events to filter by event args

ulope:
Fixes: #333


czepluch:
I can test it locally due to https://github.com/raiden-network/raiden/issues/4927, but I'd say we just merge it and rely on the scenario player to show if it doesn't work during the weekend.

nlsdfnbch:
Didnt you add tests for these previously?

PR #347, opened 2019-09-24
Refactor code for node-command parsing

nlsdfnbch:
Refactors code present in `scenario_player.node_support` into its own class.

The class handles the following jobs:

 - Creating a Keystore File
 - Creating a password File
 - Parsing `raiden` executable options and switches, and format them for usage with `subprocess`.
 - Injecting values from the scenario definition file where applicable, and handle overrides.
 - downloading and installing a raiden executable, if `raiden_version` is not  set `local` (using the `RaidenExecutable` class in `scenario_player.setup.nodes.utils`, which is stubbed for now).

Doc strings for the general setup process have been added to the module docstring of `scenario_player.setup`.

Tests have been written to cover described functionality.


nlsdfnbch:
Please review the test harness, @ulope ! Specifically, if all expectations are reasonable / make sense, and if any are missing from the original source code.

PR #357, opened 2019-10-03
Transfer Task: Accept `lock_timeout` param

rakanalh:
This is required if there is a possibility to adjust the channel's `settle_timeout` and `reveal_timeout`. The lock_timeout has to be smaller than the channel's settle timeout.


rakanalh:
Seems like "changing base branch feature" in github is broken :man_facepalming:

PR #365, opened 2019-10-19
Ensure we use the same filename for run_number

### LINKED ISSUE #364: Scenario Logger gives misleading versions, timestamps of Raiden in the node logs
Description
=========
In some of the scenarios especially the mfee scenarios the `run-000.log` file of the node logs from the SP  report that they are using raiden from 26th September 2019 or 5th October 2019 in the scenario player runs which is misleading.

```
{"raiden": "0.100.5a1.dev504+g68448ac97", "raiden_db_version": 23, "python_implementation": "CPython", "python_version": "3.7.3", "system": "Linux 64bit 4.15.0-64-generic", "architecture": "x86_64", "distribution": "source", "event": "Starting Raiden", "logger": "raiden.ui.runners", "level": "info", "timestamp": "2019-10-05 04:38:12.780377"}
```

```
{"raiden": "0.100.5a1.dev445+gf40adc88e", "raiden_db_version": 23, "python_implementation": "CPython", "python_version": "3.7.3", "system": "Linux 64bit 4.15.0-62-generic", "architecture": "x86_64", "distribution": "source", "event": "Starting Raiden", "logger": "raiden.ui.runners", "level": "info", "timestamp": "2019-09-26 06:53:06.746398"}
```
This happens almost regularly

Edit(17.10.2019)
Closer analysis of the log runs 16th Oct 2019 mfee4 scenario.
[Scenario_player_Logs-mfee4_combined_fees-1-latest-2019-10-16.tar.gz](https://github.com/raiden-network/scenario-player/files/3738952/Scenario_player_Logs-mfee4_combined_fees-1-latest-2019-10-16.tar.gz)

1. `run-000.log` of `node_0_003`
* The raiden version **"raiden": "0.100.5a1.dev445+gf40adc88e"**
* The timestamps occuring in the entire file  are **"timestamp": "2019-09-26**

2. `run-000.stdout` of `node_0_003`
* The raiden version `Welcome to Raiden, version ` **`0.100.5a1.dev445+gf40adc88e!`** same as that in `run-000.log`

3. `scenario-player-run_mfee4_combined_fees_2019-10-16T05:39:05.log`
* It gives the correct date of `2019-10-16`.
`{"version": "local", "event": "Using Raiden version", "logger": "scenario_player.node_support", "level": "info",` **`"timestamp": "2019-10-16 05:39:06.802424"}`**
* `timestamp` throughout are log file are correct for that particular date
`"timestamp": "2019-10-16`

Most probably logger picks up wrong date and version for raiden.

Expected Behavior
===============
Scenario player logger should each time notify correct raiden version and correct timestamp in all the log files that it creates.

konradkonrad:
Fix #364

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -10,21 +10,20 @@ import gevent
 import structlog
 from eth_typing import ChecksumAddress
 from eth_utils import encode_hex, is_checksum_address, to_checksum_address
-from raiden_contracts.contract_manager import ContractManager, contracts_precompiled_path
-from requests import HTTPError, RequestException, Session
-from web3 import HTTPProvider, Web3
-
 from raiden.accounts import Account
 from raiden.constants import GAS_LIMIT_FOR_TOKEN_CONTRACT_CALL
 from raiden.network.rpc.client import JSONRPCClient
 from raiden.network.rpc.smartcontract_proxy import ContractProxy
 from raiden.utils.typing import TransactionHash
+from raiden_contracts.contract_manager import ContractManager, contracts_precompiled_path
+from requests import HTTPError, RequestException, Session
 from scenario_player.constants import (
     API_URL_TOKEN_NETWORK_ADDRESS,
     API_URL_TOKENS,
     NODE_ACCOUNT_BALANCE_FUND,
     NODE_ACCOUNT_BALANCE_MIN,
     OWN_ACCOUNT_BALANCE_MIN,
+    RUN_NUMBER_FILENAME,
 )
 from scenario_player.exceptions import ScenarioError, TokenRegistrationError
 from scenario_player.exceptions.legacy import TokenNetworkDiscoveryTimeout
@@ -33,6 +32,7 @@ from scenario_player.services.rpc.utils import assign_rpc_instance_id
 from scenario_player.services.utils.interface import ServiceInterface
 from scenario_player.utils import TimeOutHTTPAdapter, get_udc_and_token, wait_for_txs
 from scenario_player.utils.token import Token, UserDepositContract
+from web3 import HTTPProvider, Web3
 
 if TYPE_CHECKING:
     from scenario_player.tasks.base import Task, TaskState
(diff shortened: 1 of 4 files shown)


scenario_player/runner.py
```diff
@@ -10,21 +10,20 @@
...
-from web3 import HTTPProvider, Web3
-
 from raiden.accounts import Account
 from raiden.constants import GAS_LIMIT_FOR_TOKEN_CONTRACT_CALL
 from raiden.network.rpc.client import JSONRPCClient
 from raiden.network.rpc.smartcontract_proxy import ContractProxy
 from raiden.utils.typing import TransactionHash
+from raiden_contracts.contract_manager import ContractManager, contracts_precompiled_path
+from requests import HTTPError, RequestException, Session
 from scenario_player.constants import (
```
nlsdfnbch: I doubt that's correct. Did you run `make style` ?

konradkonrad:
> a test that the constant did not change would be appropriate.

Not sure, what kind of test you want here? It's a constant for a filename, not for a number. So for me this is in the style and best-practices realm: never spell out the filename, but always use the constant. Enforce on review.

nlsdfnbch:
If the constant is changed, it'll break backwards-compatiblity. The run number is used to generate wallets for nodes deterministically. So changing that constant at any point in time will result in unexpected behaviour. 

While typically I would'nt request this, a test asserting the constant remains unchanged makes sense in this case IMO.  Think of it less as a unit test, and more as a sanity check.

```
def test_run_number_constant_remains_unchanged():
    assert RUN_NUMBER_FILENAME == "run_number.txt"
```

konradkonrad:
> While typically I would'nt request this, a test asserting the constant remains unchanged makes sense in this case IMO. Think of it less as a unit test, and more as a sanity check.
> 
> ```
> def test_run_number_constant_remains_unchanged():
>     assert RUN_NUMBER_FILENAME == "run_number.txt"
> ```

This isn't worth much, as any sane developer would refactor the name in both places. I added a heavy comment on top of the constant declaration. If you want more than this, I'm afraid you're on your own.

Also something I noticed: running `make style` consecutively applies changes every time, so it seems that some steps interfere with each other.

Issue #350, opened 2019-09-27
Develop a smoke test for the scenario player


christianbrb:
Analysis
=======
Current State
-----------------
Currently, the test coverage is at ~60%. It is important to keep the scenario player stable. Therefore we are running a manual test if major changes have been made.

Problem
-----------
The manual tests are time-consuming and might not touch all parts of the software. In addition small changes might not get smoke tested and therefore lead to errors.

Solution
----------
Create a pragmatic way (eg. running the bf1 scenario) for a smoke test

nlsdfnbch:
Running one of the nightly scenarios seems to touch on too many components that would be out of scope for a smoke test.

Per definition, smoke tests should verify the build, indicating if any further testing makes sense.

That being said, I believe the following would be the desired scope of such tests:

- Node Setup works (we can fire up a number of raiden nodes for a scenario) - this includes looking up binaries, creating keystores and starting the processes
- Node funding works
- udc deposits work
- deploying custom tokens works
- configuring a pfs works
- tear down works

Luckily, this can be tested with a no-op scenario. Although this isn't quite ideal (it still requires an internet connection for contract deployments, etc), it would serve as a good start.

nlsdfnbch:
The linked PR now extends to including a utility, as we cannot test the previously stated features without shipping a funded account (or requiring the user to supply one).

The utility creates a complete stack to run a scenario - test net genesis file, eth nodes, and contracts.

It currently relies on an update to raiden to include a new `raiden_contracts` version, as `0.25.1` is currently undeployable to private networks.

nlsdfnbch:
@Dominik1999 please fund `0x54b819488D4c864c4098BCA61F1BF840E52D2f38` with some goeth for our smoketest scenario.

Dominik1999:
https://goerli.etherscan.io/tx/0xbf2f814f957da4ed240411d3d737be88c32a7b7e7d5f5c046763bea618b5d12b

nlsdfnbch:
https://circleci.com/gh/raiden-network/scenario-player/2662

Smoketests are run successfully on circle.

PR #366, opened 2019-10-21
Add smoketest utility

nlsdfnbch:
Implements #350 .

Touches the following features, testing their defaults:

- [x] Node Setup works (we can fire up a number of raiden nodes for a scenario)
- [x] Node funding works
- [x] udc deposits work
- [x] deploying custom tokens works
- [x] configuring a pfs works
- [x] tear down works (exit code 0)


nlsdfnbch:
This is meant as a smoketest for the SP - it should confirm most parts of it are working.
I just wanted a pair of eyes to confirm that the intention is clear. 

I've updated the yaml file with a doc string to clear up its purpose.

The only thing I see that isn't tested is the funding of nodes, since we can't provide a stocked wallet with the SP.

nlsdfnbch:
I've updated the PR. This should be taken with a grain of salt, as I think it's quite a lot of staffolding for a sanity check, but alas I suppose it comes with the design of the software.

There now exists a python script at `tests/smoketests/smoketest_launcher.py` which will deploy a custom network, including raiden contracts. It then creates a scenario definiion file from a jinja2 template and runs that scenario.

Finally, it cleans up any generated files.

@yoichi, please have a look at the contract deployment part.

pirapira:
Until the contract fix lands, deployments will fail: https://github.com/raiden-network/raiden-contracts/pull/1313

nlsdfnbch:
After a chat with @konradkonrad , we found a simpler solution to spinning up a private network. An account file has been supplied, which is unlocked using a circleci secret. 

Circle will run the smoke tests against PRs and bumps, to verify the build.

This PRs introduces the changes necessary to make this happen.

PR #367, opened 2019-10-23
Correcting the example scenario because v1 does not exist anymore

### LINKED ISSUE #360: Remove external node configuration to example file
## Problem

Within the current example file it is written:

```yaml
nodes:
  ## Mode, either `managed` or `external`
  # - managed: The scenario player will manage the Raiden nodes
  # - external: The nodes need to be managed externally (see v1 for examples)
  mode: managed
```
The v1 example file is not avaible anymore in 
https://github.com/raiden-network/scenario-player/tree/master/examples

The commit which deleted it was
https://github.com/raiden-network/scenario-player/commit/d9390849f46abbe57173106a04a555e776432546#diff-b7e6ad2834c284e1323007cada49bddf

## Solution

Remove in [the current example file](https://github.com/raiden-network/scenario-player/blob/master/examples/scenario-example-v2.yaml) the instructions how to manage an external node.

agatsoh:
fixes #360

### DIFF
--- a/examples/scenario-example-v2.yaml
+++ b/examples/scenario-example-v2.yaml
@@ -53,9 +53,6 @@ token:
   #reuse: true
 
 nodes:
-  ## Mode, either `managed` or `external`
-  # - managed: The scenario player will manage the Raiden nodes
-  # - external: The nodes need to be managed externally
   mode: managed
   count: 3
   ## Raiden version to use, default: 'LATEST' uses the latest nightly release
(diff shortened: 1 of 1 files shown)


agatsoh:
Removed all the lines mentioning mode

agatsoh:
> ..including `mode: managed` smile
Done . But i observed that it is present  in almost all the scenarios that we currently have. So somebody needs to bring it to notice.

nlsdfnbch:
> > ..including `mode: managed` smile
> 
> Done . But i observed that it is present in almost all the scenarios that we currently have. So somebody needs to bring it to notice.

It's a deprecated option - it doesn't hurt to state it, but it's also not necessary. Feel free to open an issue for the options removal!

PR #368, opened 2019-10-24
Update Dockerfile

### LINKED ISSUE #363: Dockerize the scenario player
Analysis
=======
Current State
-----------------
We currently pull the raiden and SP repo grom github via a bash script, build a docker container and use it in the nightlyies, every night.

Problem
-----------
The bash script is delicate and not exactly robust, the process of building a docker container is
additional load which can be easily offloaded to dockerhub.

Solution
----------
Write a new docker setup for the scenario player

nlsdfnbch:
Fixes #363

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,62 @@
-# Set the python version to use. This may be overriden during the build process by passing
-# --build-arg PY_VERSION=<desired version>
-# Note that the desired version must be of pattern <major>.<minor> .
-ARG PY_VERSION=3.7
-FROM python:$PY_VERSION AS cache
+FROM python:3.7-slim AS compile-image
 
-RUN pip install raiden
+LABEL maintainer=devops@brainbot.com
 
-FROM python:${PY_VERSION}
+################################################################################
+# COMPILATION STAGE (compile-image)                                            #
+################################################################################
+#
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,22 +1,62 @@
...
+#
+# Copies the raiden client, scenario player and scenario defnition files from the
+# compilation stage, and configure further environment variables required to run
+# the nightlies.
+#
+# This stage copies the virtual environment created in the previous stage,
+# and prepends its location to the $PATH environment variable.
+
+# Copy virtual env and scenario definition files.
+COPY --from=compile-image /opt/venv /opt/venv
```
konradkonrad: Depending on how you intend to use this, it would make sense to copy the scenarios from the raiden checkout above.

konradkonrad: The alternative is of course to mount the scenario descriptions into the running container (which will need another git checkout step previous to the execution)

nlsdfnbch:
Scenarios are now part of the container

Dockerfile
nlsdfnbch: Scenarios are now part of the container

nlsdfnbch:
Yepp, unrelated

PR #372, opened 2019-10-25
Support unknown flags and options

nlsdfnbch:
This should no longer limit the CLI options supported by the SP.

Flags which are not present in `FLAG_OPTIONS` can be passed by setting their value to a `True` or `False`
```yaml
nodes:
  default-options:
    this_is_a_flag_option: true
    address: 1234
```

Results in the raiden client being run like so
```bash
raiden --this_is_a_flag_option --address 1234 [...]
```

Since this is a cherry-pick I'm not including tests, those will be merged once the refac Pr (from which this stems) is done.

### DIFF
--- a/scenario_player/node_support.py
+++ b/scenario_player/node_support.py
@@ -72,6 +72,7 @@ MANAGED_CONFIG_OPTIONS_OVERRIDABLE = {
 
 KNOWN_OPTIONS = {param.name.replace("_", "-") for param in cli_run.params}
 FLAG_OPTIONS = {param.name.replace("_", "-") for param in cli_run.params if param.is_flag}
+FLAG_OPTIONS += {'no-' + opt for opt in FLAG_OPTIONS}
 
 
 class NodeState(Enum):
@@ -318,7 +319,10 @@ class NodeRunner:
         if pfs_address:
             cmd.extend(["--pathfinding-service-address", pfs_address])
 
+        seen = set()
+
         for option_name in MANAGED_CONFIG_OPTIONS_OVERRIDABLE:
+            seen.add(option_name)
             if option_name in ("api-address", "pathfinding-service-address"):
                 # already handled above
                 continue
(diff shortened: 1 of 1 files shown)


scenario_player/node_support.py
```diff
@@ -343,6 +347,15 @@ def _command(self):
                 else:
                     cmd.append(f"--{option_name}")
 
+        for option_name in self._options.keys() - seen:
+            option_value = self._options[option_name]
+            if isinstance(option_value, list):
+                cmd.extend([f"--{option_name}", *self._options[option_name]])
+            else:
+                cmd.extend([f"--{option_name}", self._options[option_name]])
```
karlb: Flags will run in this case, too. So you will get both `--no-cap-mediation-fees True` and `--no-cap-mediation-fees ` as args.

scenario_player/node_support.py
```diff
@@ -343,6 +347,15 @@ def _command(self):
                 else:
                     cmd.append(f"--{option_name}")
```
karlb: Doesn't this already handle positive flags properly? It looks like we could replace this PR by
```python
FLAG_OPTIONS += {'no-' + opt for opt in FLAG_OPTIONS}
```
so that negative options are properly handled, too.

nlsdfnbch: Ach, I got `FLAG_OPTIONS` confused with the pre-defined lists in the `MANAGED_*` constants. 
You are correct, that would work just fine.

nlsdfnbch:
I'll close this PR since it has turned out to be a `raiden` issue, not an SP issue:

```
from raiden.ui.cli import run as cli_run
...
FLAG_OPTIONS = {param.name.replace("_", "-") for param in cli_run.params if param.is_flag}
```

Valid flag options are read from the `raiden` client (I falsely believed this to be a pre-defined list in the SP  - blame a tired mind for not spotting this earlier).

karlb:
I would still expect this:
```python
FLAG_OPTIONS += {'no-' + opt for opt in FLAG_OPTIONS}
```
to be in the scenario player, since having a single flask `option` for a boolean flag is correct, even though it can take two different forms on the CLI.

Issue #399, opened 2019-11-06
Support TransactionSlot


nlsdfnbch:
https://github.com/raiden-network/raiden/commit/8c5285afa64a41f1af70cdac815b4962a17aeb64#diff-79ef917c2473698777a5474a23dc119f introduced a new interface for sending transactions via the `JSONRPCClient`.

As a result, we'll have to update the RPC service accordingly, in order to avoid `500` crashes.

This includes updating the tests using a mocked RPC client instance.

nlsdfnbch:
Back from the dead, this also break the `reclaim-eth` function in `scenario_player.utils.legacy` - A fix is required for this as well.

PR #400, opened 2019-11-06
[FIX-#399] Support TransactionSlots

nlsdfnbch:
Introduces a cusom `send_transaction` method on our RPCClient class, which takes the necessary steps required by the raiden JSONRPCClient under the hood.

### DIFF
--- a/scenario_player/services/rpc/utils.py
+++ b/scenario_player/services/rpc/utils.py
@@ -43,6 +43,10 @@ class RPCClient(JSONRPCClient):
         )
         self.client_id = generate_hash_key(chain_url, privkey, strategy)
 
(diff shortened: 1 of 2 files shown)


scenario_player/services/rpc/utils.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, chain_url, privkey, strategy):
         )
         self.client_id = generate_hash_key(chain_url, privkey, strategy)
 
+    def send_transaction(self, to, startgas: int, value: int = 0, data: bytes = b""):
+        transaction_slot = super(RPCClient, self).get_next_transaction()
+        return transaction_slot.send_transaction(to, startgas, value, data)
```
Dominik1999: you might want to write a test for that?

Issue #334, opened 2019-09-16
Make The `ScenarioYAML` object the single source of truth for the scenario configuration


nlsdfnbch:
Analysis
=======
Current State
-----------------
Currently, a lot of the settings are stored in the scenario definition file (`.yaml`), but some are passed via CLI (`--chain`) and others are hard-coded into Constants.

Problem
-----------
Passing the cli flags around and accessing the constants makes for busy function signatures and sometimes pretty annoying code.

Solution
----------
Load all CLI flags and constants used into the `ScenarioYAML` instance which is loaded anyway at the beginning of each run, and pass this around instead. It provides a consistent user interface, and would improve clarity on what is handed to a part of the Scenario Player code.

PR #401, opened 2019-11-06
[FEAT-#334] Store scenario settings in ScenarioDefinition only

nlsdfnbch:
Stores all scenario-related options in the ScenarioDefinition
object or a mapping it contains.

Makes --chain optional, and only allows one instance of it to be
passed. Default chain is `goerli`, default node is
 `parity.goerli.ethnodes.brainbot.com:8545`.

Rationale behind making `--chain` optional is that passing it multiple times is not used in prod, nor is using the option `chain: any` in the scenario definitions.

~I'm currently still running into issues with the CLI tests, which seem to fail due to some unpacking error.~

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -5,7 +5,6 @@ import json
 import os
 import sys
 import traceback
-from collections import defaultdict
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
@@ -21,6 +20,7 @@ from raiden.accounts import Account
 from raiden.log_config import _FIRST_PARTY_PACKAGES, configure_logging
 from raiden.utils.cli import EnumChoiceType
 from scenario_player import __version__, tasks
+from scenario_player.constants import DEFAULT_ETH_RPC_ADDRESS, DEFAULT_NETWORK
 from scenario_player.exceptions import ScenarioAssertionError, ScenarioError
 from scenario_player.exceptions.cli import WrongPassword
 from scenario_player.exceptions.services import ServiceProcessException
@@ -28,12 +28,7 @@ from scenario_player.runner import ScenarioRunner
 from scenario_player.services.common.app import ServiceProcess
 from scenario_player.tasks.base import collect_tasks
 from scenario_player.ui import ScenarioUI, attach_urwid_logbuffer
-from scenario_player.utils import (
-    ChainConfigType,
-    DummyStream,
-    post_task_state_to_rc,
-    send_notification_mail,
-)
+from scenario_player.utils import DummyStream, post_task_state_to_rc, send_notification_mail
 from scenario_player.utils.legacy import MutuallyExclusiveOption
 from scenario_player.utils.version import get_complete_spec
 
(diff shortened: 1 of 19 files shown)


scenario_player/runner.py
```diff
@@ -66,36 +65,29 @@ def __init__(
...
         log.debug("Local seed", seed=self.local_seed)
 
-        self.data_path = self.base_path.joinpath("scenarios", scenario_name)
-        self.data_path.mkdir(exist_ok=True, parents=True)
-
-        self.definition = ScenarioDefinition(scenario_file, self.data_path)
-        log.debug("Data path", path=self.data_path)
-
-        # Determining the run number requires :attr:`.data_path`
+        # Determining the run number requires :attr:`.yaml`
```
konradkonrad: What is this comment supposed to tell me?

scenario_player/runner.py
```diff
@@ -45,10 +44,10 @@ class ScenarioRunner:
     def __init__(
         self,
         account: Account,
-        chain_urls: Dict[str, List[str]],
         auth: str,
         data_path: Path,
         scenario_file: Path,
+        chain_url: Optional[str] = None,
```
konradkonrad: how is this optional?

tests/unittests/cli/test_cli.py
```diff
@@ -13,7 +13,6 @@
 KEYSTORE_PATH = Path(__file__).resolve().parent.joinpath("keystore")
 SCENARIO = f"{Path(__file__).parent.joinpath('scenario', 'join-network-scenario-J1.yaml')}"
 CLI_ARGS = (
-    f"--chain goerli:http://geth.goerli.ethnodes.brainbot.com:8545 "
```
konradkonrad: please add a unit test that ensures the cli flag is still functional and respected

konradkonrad: I see, that you added https://github.com/raiden-network/scenario-player/pull/401/files#diff-682d38b401009dead7fde5128615d0b5R89 ( `test_chain_property_prioritizes_cli_attr_over_scenario_definition` ) -- but I'm afraid this is still prone to later regressions.

scenario_player/runner.py
konradkonrad:
```suggestion
        # Determining the run number requires a scenario definition from `yaml`.
```

scenario_player/runner.py
konradkonrad: I got it, there is a default now.

scenario_player/main.py
```diff
@@ -156,10 +156,9 @@ def chain_option(func):
     @click.option(
         "--chain",
         "chains",
-        type=ChainConfigType(),
-        multiple=True,
-        required=True,
-        help="Chain name to eth rpc url mapping, multiple allowed",
+        multiple=False,
+        required=False,
```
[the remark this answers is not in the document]
konradkonrad: I see, you introduced a default...

scenario_player/runner.py
nlsdfnbch: it's supposed to tell you that you need to load the scenario definition first, before you can determine the run number. It just does a terrible job.

nlsdfnbch:
I conclude that the smoketest failure is a caching problem and utter horseshit - the reason this fails is because of a bug with the CLI options, which was already fixed and isn't even present any longer in the code.

PR #404, opened 2019-11-14
Remove pack-logs command and related code.

nlsdfnbch:


christianbrb:
How would I access the logs after this has been removed?

nlsdfnbch:
via port 8000 of the SP instance, as always.

PR #437, opened 2019-12-03
Pass service port to SPAAS service

rakanalh:
Introduce a new CLI param to be able to specify the port
for the SPAAS service

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -24,6 +24,7 @@ from scenario_player.constants import DEFAULT_ETH_RPC_ADDRESS, DEFAULT_NETWORK
 from scenario_player.exceptions import ScenarioAssertionError, ScenarioError
 from scenario_player.exceptions.cli import WrongPassword
 from scenario_player.exceptions.services import ServiceProcessException
+from scenario_player.definition import ScenarioDefinition
 from scenario_player.runner import ScenarioRunner
 from scenario_player.services.common.app import ServiceProcess
 from scenario_player.tasks.base import collect_tasks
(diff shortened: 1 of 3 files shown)


rakanalh:
The reason we're doing this is that a `Address In Use` exception is raised when running multiple SPs on the same machine. Regardless of the exception, the SPs continue to run the scenarios as expected.

The PR is a PoC on trying to prevent raising an exception. However, as you're suggesting a different approach, i wouldn't mind closing this PR and implementing this at some point.

scenario_player/main.py
```diff
@@ -228,18 +236,27 @@ def run(
...
+        }
+    }
+    scenario_definition = ScenarioDefinition(scenario_file, data_path, overrides)
+
     # Dynamically import valid Task classes from sceanrio_player.tasks package.
     collect_tasks(tasks)
 
     # Start our Services
-    service_process = ServiceProcess()
+    service_process = ServiceProcess(port=service_port)
```
rakanalh: @nlsdfnbch Not sure i get your comment. You mean i should be using the `scenario_definition.spaas.rpc.port` here?

nlsdfnbch: Yupp, that's what I'm saying.

nlsdfnbch:
> The reason we're doing this is that a `Address In Use` exception is raised when running multiple SPs on the same machine. Regardless of the exception, the SPs continue to run the scenarios as expected.
> 
> The PR is a PoC on trying to prevent raising an exception. However, as you're suggesting a different approach, i wouldn't mind closing this PR and implementing this at some point.

Unless you have dedicated accounts for each run of the sp, running them concurrently wouldn't work anyway. 

If you do have dedicated accounts, you could alternatively run them in a docker container, which should work.

scenario_player/main.py
rakanalh: Done.

rakanalh:
btw, there's a false positive on CI because a lot of unit tests are failing for me locally.

PR #439, opened 2019-12-04
Do not treat GreenletExit as error.

nlsdfnbch:


nlsdfnbch:
@agatsoh please test if the `wait` tasks are still failing. Specifically, check the following (preferably in a scenario each)

- [ ] executing a single, successful `wait` task within `serial` does *not* raise an error.
- [ ] executing a single `wait` task and another task of your choice (the task must **succeed**) within `serial` does not raise an error.
- [ ] executing a single `wait` task and another task of your choice (the task must **fail**) within `serial` does not raise a GreenletExit error, but an appropriate TaskError for the failing task.

agatsoh:
@nlsdfnbch The `bf2_long_running` scenario has all the conditions that you specified above. So not creating any extra scenario just running the `bf2_long_running` which would help us test the PR.

agatsoh:
@nlsdfnbch The scenario was successfull. Do you think this change would be fit to be merged ?

nlsdfnbch:
I should hope so. Let's see how does outside of the lab.

PR #441, opened 2019-12-05
Add retries for AssertTask

palango:
@Dominik1999 The stuff we talked about.

Open question is how big we should make the timeout. Any opinions?


Dominik1999:
nice merge it ! merge it!

PR #453, opened 2019-12-10
Move retry logic to base Task

palango:
Use the retry logic in the other asserts as well.

### DIFF
--- a/scenario_player/tasks/base.py
+++ b/scenario_player/tasks/base.py
@@ -82,8 +82,8 @@ class Task:
     # tasks that are side-effect free can overwrite the value to 0.
     #
     # Ref.: https://github.com/raiden-network/raiden/issues/6149#issuecomment-627387624
-    SYNCHRONIZATION_TIME_SECONDS = 30
-    DEFAULT_TIMEOUT = 0  # 5 * 60  # 5 minutes
+    SYNCHRONIZATION_TIME_SECONDS = 0  # keep the code for now, can be removed in the future
+    DEFAULT_TIMEOUT = 0  # Tasks that need retries need to overwrite this
 
     def __init__(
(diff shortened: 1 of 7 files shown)


scenario_player/tasks/base.py
```diff
@@ -82,6 +83,7 @@ class Task:
     #
     # Ref.: https://github.com/raiden-network/raiden/issues/6149#issuecomment-627387624
     SYNCHRONIZATION_TIME_SECONDS = 30
+    DEFAULT_TIMEOUT = 0  # 5 * 60  # 5 minutes
```
ulope, line 86: `0 == 5 minutes`? ;)

scenario_player/tasks/base.py
```diff
@@ -82,6 +83,7 @@ class Task:
     #
     # Ref.: https://github.com/raiden-network/raiden/issues/6149#issuecomment-627387624
     SYNCHRONIZATION_TIME_SECONDS = 30
```
ulope, line 85: Do we still need this?

scenario_player/tasks/base.py
```diff
@@ -108,8 +110,28 @@ def __call__(self, *args, **kwargs):
...
+                            except ScenarioAssertionError as ex:
+                                exception = ex
+
+                            if return_val:
+                                return return_val
+
+                            sleep(1)
+                except Timeout:
+                    if exception:
+                        raise exception
```
ulope, line 132: It would probably be good to add a bit of logging here (e.g. retry count, maybe remaining timeout...)

scenario_player/tasks/base.py
palango: I set it to 0 for now, once the scenarios run fine I'm going to remove it.

PR #454, opened 2019-12-12
Fix #450

### LINKED ISSUE #450: Proper logging for node startup failures
check https://github.com/raiden-network/raiden/issues/5459

nlsdfnbch:


scenario_player/node_support.py
```diff
@@ -510,9 +510,13 @@ def _monitor(runner: NodeRunner):
                     try:
                         runner.executor.check_subprocess()
                     except ProcessExitedWithError as ex:
+                        log.error("Node Error - Process", exception=ex)
```
ulope: I'd use `log.exception()` that way the entire exception context will be logged.

nlsdfnbch: ah, my IDE didnt autocomplete on that, so I thought it wasnt available in `structlog`.

PR #467, opened 2020-01-13
Fix node state tracking

ulope:
There was a race between the node monitor and the `stop()` / `kill()` methods.
The node state was only updated after they were stopped, which could lead to random failures.


ulope:
That would be nice. Unfortunately there are currently no tests at all for this part of the SP...

Dominik1999:
Merging now for tonights test. @nlsdfnbch should take another look tomorrow and maybe add a test?

PR #475, opened 2020-02-03
Minting didn't handle chain pos. reorgs

Dominik1999:
Might have caused #5802 - tests are missing

How could we test this?

### DIFF
--- a/scenario_player/utils/legacy.py
+++ b/scenario_player/utils/legacy.py
@@ -16,7 +16,7 @@ import mirakuru
 import requests
 import structlog
 from eth_keyfile import decode_keyfile_json
-from eth_utils import to_checksum_address
+from eth_utils import to_checksum_address, remove_0x_prefix
 from mirakuru import AlreadyRunning, TimeoutExpired
 from mirakuru.base import ENV_UUID, IGNORED_ERROR_CODES
 from raiden_contracts.constants import CONTRACT_CUSTOM_TOKEN, CONTRACT_USER_DEPOSIT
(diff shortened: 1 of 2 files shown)


scenario_player/utils/legacy.py
```diff
@@ -197,35 +197,36 @@ def _set_timeout(self, timeout=None):
...
     start = time.monotonic()
     outstanding = False
-    txhashes = set(txhashes)
-    while txhashes and time.monotonic() - start < timeout:
+    txhashes_bytes = []
+
+    ## for some reason txhashes is a set of str, not bytes
+    for txhash in txhashes:
+        if isinstance(txhash, str):
+            txhashes_bytes.append(bytes.fromhex(remove_0x_prefix(txhash)))
```
ulope: As discussed in person we should fix this at the source.

scenario_player/utils/legacy.py
```diff
@@ -197,35 +197,36 @@ def _set_timeout(self, timeout=None):
...
+            outstanding = len(txhashes_bytes)
             log.debug(
                 "Waiting for tx confirmations",
                 outstanding=outstanding,
                 timeout_remaining=int(remaining_timeout),
             )
-        for txhash in txhashes.copy():
-            tx = web3.eth.getTransactionReceipt(txhash)
+        for txhash in txhashes_bytes.copy():
+            tx = client.poll(txhash)
```
ulope: This probably fixes the issue, but is suboptimal since it blocks on each tx until they are mined.

Instead I'd suggest to add the logic from poll in here.
Basically that's only adding a check that the block number returned from `eth.getTransactionReceipt` is `current_block` + `confirmation_blocks`.

scenario_player/utils/legacy.py
Dominik1999: done, I now fixed that at the source directly to convert any str into bytes

Dominik1999:
creating a new PR

scenario_player/utils/legacy.py
Dominik1999: this will be covered here https://github.com/raiden-network/scenario-player/issues/476

PR #477, opened 2020-02-04
Minting didn't handle chain pos. reorgs

Dominik1999:
Might have caused #5802 - tests are missing

How could we test this?

### DIFF
--- a/scenario_player/utils/legacy.py
+++ b/scenario_player/utils/legacy.py
@@ -223,7 +223,7 @@ def wait_for_txs(
                 if status is not None and status == 0:
                     raise ScenarioTxError(f"Transaction {txhash} failed.")
                 ## we want to add 2 blocks as confirmation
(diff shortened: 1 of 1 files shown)


scenario_player/utils/legacy.py
```diff
@@ -222,6 +222,9 @@ def wait_for_txs(
                 status = tx.get("status")
                 if status is not None and status == 0:
                     raise ScenarioTxError(f"Transaction {txhash} failed.")
+                ## we want to add 2 blocks as confirmation
+                if tx["blockNumber"] + 2 >= web3.eth.getBlock("latest")['number']:
+                    continue
                 txhashes.remove(txhash)
```
ulope: I would invert the condition and move the `txhashes.remove()` call into the condition. 
As it's currently the `continue` will also skip the `sleep` and thereby spam the eth node with requests.

Dominik1999: true

Issue #474, opened 2020-01-30
Don't report nodes killed by the scenario player as failures


hackaugusto:
Do this after https://github.com/raiden-network/scenario-player/issues/468.

Under some circumstances a node will die and the scenario player will stop the run, if the stop takes longer than expected a SIGKILL is sent. When this happens the scenario player should not report the death of the node as a failure, since that is masquerading the real problem.

Dominik1999:
So we want the scenario player to continue the run even after is has sent a SIGKILL to a node. So the SP should log that killing, but continue to run the scenario.

Dominik1999:
@ulope

hackaugusto:
> So we want the scenario player to continue the run even after is has sent a SIGKILL to a node

This is the case, the problem is that the scenario player wrongly reports the kill it has done as an error.

Dominik1999:
Ulo will adjust the code https://github.com/raiden-network/scenario-player/blob/a8e8c16bb0b010289f2267cb667b44cb3149f926/scenario_player/utils/legacy.py#L187

such that this error is not handled as a failure

Dominik1999:
- Increase timeout for stop mirakuru to 10 min

Dominik1999:
Are you working on this issue atm, @konradkonrad ?

Dominik1999:
It looks like atm the Scenario Player does not report a node death as a failure

here http://scenario-player.ci.raiden.network/scenarios/bf1_basic_functionality/scenario-player-run_bf1_basic_functionality_2020-02-25T00%3A29%3A30.log.gz

we see `Node error` of Node 1 at `2020-02-25 01:40:24`

whereas the scenario continues working 5 minutes more until `2020-02-25 01:40:24.930022`

Dominik1999:
This issue seems not to be solved. Tonight all scenarios failed because of the same reason, the pfs was not online. But some of the nodes stopped by 1 and other by 9. So the Scenario Player does seem to still report a node being killed (that indicates the -9) as a failure.

``` 
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Scenario scenario-player-run_pfs5_too_low_capacity_2020-03-04T21:21:29.log.gz failed
[...]

PR #479, opened 2020-02-11
Increase node stop timeout to 10mins

palango:
As discussed in https://github.com/raiden-network/scenario-player/issues/474#issuecomment-583005587


nlsdfnbch:
Instead you guys should just implement an API endpoint to check node state, and change the assertion method to requesting this endpoint. /endofbickering

@palango can you do me a favor and also remove the codeowner file? you guys don't need my reviews for these.

PR #489, opened 2020-02-20
Simplify Circle config

palango:
So the tests did actually never run.

- Simplify CircleCI config
- Remove commit message checks
- Skip failing tests: https://github.com/raiden-network/scenario-player/issues/490
- Skip smoketests


.circleci/config.yml
```diff
@@ -223,72 +200,28 @@ jobs:
...
-            ssh-keyscan -H github.com >> /home/circleci/.ssh/known_hosts
-      - run:
-          name: install project.
-          command: pip install ${PROJECT_ROOT}
-
-      - run:
-          name: Bump part on commits.
-          command: |
-            python3 ${CI_SCRIPTS_DIR}/bump-part.py
+      # - run_smoketests
```
karlb: Why don't we run the smoketests?

palango: They currently fail as well: https://app.circleci.com/jobs/github/raiden-network/scenario-player/3625

So I think it's better to merge this as it and fix them in a follow-up.

palango: I'll create an issue

PR #501, opened 2020-02-28
Restore optional account reuse + Snapshot support

ulope:
This adds the following features:
- Re-enables the possibility to reuse node accounts between runs (as a per scenario option)
- Adds a "snapshot" feature that allows to restore the node datadirs to a previous state. 

Additionally this fixes the TUI that had been broken by a previous refactoring.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,14 +21,14 @@ requires = [
     'click',
     'eth-utils',
     'gevent',
-    'pyyaml',
-    'requests',
-    'raiden',
-    'structlog',
-    'simplejson',
-    'urwid!=2.1.0',
-    'waitress',
     'jinja2',
+    'prometheus_client<0.8.0',
+    'pyyaml',
+    'raiden',
+    'requests',
+    'simplejson',
+    'structlog',
+    'urwid!=2.1.0',
 ]
 
 # Classifiers, Licensing, keywords for pypi.
@@ -53,7 +53,7 @@ keywords = "raiden, raiden-network"
 [tool.flit.metadata.requires-extra]
 # Extended dependencies required for development and testing. Flit installs
 # these automatically.
-test = ["pytest==4.5.0"]
+test = ["pytest"]
 dev = [
     "black==19.3b0",
     "coverage",
@@ -64,7 +64,7 @@ dev = [
     "isort==4.3.16",
     "matrix-synapse==1.10.1",
     "mypy==0.761",
-    "pytest",
+    "pylint",
     "pytest-cov",
     "pytest-dependency==0.4.0",
     "responses==0.10.6",
--- a/scenario_player/definition.py
+++ b/scenario_player/definition.py
@@ -23,6 +23,7 @@ class ScenarioDefinition:
     def __init__(
         self, yaml_path: pathlib.Path, data_path: pathlib.Path, environment: EnvironmentConfig
     ) -> None:
+        self._scenario_dir = None
         self.path = yaml_path
         # Use the scenario file as jinja template and only parse the yaml, afterwards.
         with yaml_path.open() as f:
@@ -30,12 +31,11 @@ class ScenarioDefinition:
             rendered_yaml = yaml_template.render(**asdict(environment))
             self._loaded = yaml.safe_load(rendered_yaml)
 
-        self._scenario_dir = None
-        self.token = TokenConfig(self._loaded, data_path.joinpath("token.info"))
-        deploy_token = self.token.address is None
-        self.nodes = NodesConfig(self._loaded, environment="development" if deploy_token else None)
         self.settings = SettingsConfig(self._loaded, environment)
         self.settings.sp_root_dir = data_path
+        self.token = TokenConfig(self._loaded, self.scenario_dir.joinpath("token.info"))
+        deploy_token = self.token.address is None
+        self.nodes = NodesConfig(self._loaded, environment="development" if deploy_token else None)
         self.scenario = ScenarioConfig(self._loaded)
 
         # If the environment sets a list of matrix servers, the nodes must not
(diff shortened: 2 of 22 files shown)


pyproject.toml
```diff
@@ -20,15 +20,21 @@ requires-python = ">3.7"
 requires = [
     'click',
     'eth-utils',
+    'eth-utils',
```
palango, line 23: This is double

pyproject.toml
```diff
@@ -20,15 +20,21 @@ requires-python = ">3.7"
...
+    'flask==1.1.2',
```
palango, line 24: why pin this?

scenario_player/definition.py
```diff
@@ -23,19 +23,19 @@ class ScenarioDefinition:
...
             rendered_yaml = yaml_template.render(**asdict(environment))
             self._loaded = yaml.safe_load(rendered_yaml)
 
-        self._scenario_dir = None
-        self.token = TokenConfig(self._loaded, data_path.joinpath("token.info"))
-        deploy_token = self.token.address is None
-        self.nodes = NodesConfig(self._loaded, environment="development" if deploy_token else None)
         self.settings = SettingsConfig(self._loaded, environment)
         self.settings.sp_root_dir = data_path
+        self.token = TokenConfig(self._loaded, self.scenario_dir.joinpath("token.info"))
```
palango, line 36: How can `self._scenario_dir` not be `None` here?

pyproject.toml
ulope: Argh, slipped through the rebase.

pyproject.toml
ulope: Why have it at all is also a good question. I think I may have botched the rebase for the dependencies here.
There is no longer an API server etc. in the SP...

scenario_player/definition.py
ulope: `scenario_dir` (without the underscore) is a property. It reads it from the settings.

scenario_player/definition.py
```diff
@@ -50,9 +50,15 @@ def name(self) -> str:
         return self.path.stem
 
     @property
-    def scenario_dir(self):
+    def scenario_dir(self) -> pathlib.Path:
```
palango, line 53: 🎖️  for adding annotations

scenario_player/main.py
```diff
@@ -268,20 +272,23 @@ def _load_environment(environment_file: IO) -> EnvironmentConfig:
...
-    keystore_file,
-    scenario_file,
-    notify_tasks,
-    enable_ui,
-    password_file,
-    log_file_name,
+    data_path: Path,
+    auth: Optional[str],
+    password: Optional[str],
+    keystore_file: str,
```
palango, line 284: Should we make all files to `Path`s?

scenario_player/tasks/execution.py
```diff
@@ -55,10 +55,48 @@ class ParallelTask(SerialTask):
     _name = "parallel"
 
     def _run(self, *args, **kwargs):
-        group = Group()
+        pool = Pool(size=self._config.get("concurrency", None))
```
palango, line 58: Is there any usecase for a limited pool?

scenario_player/utils/configuration/nodes.py
```diff
@@ -71,12 +81,35 @@ def validate(self):
 
             * The scenario version is > 1
             * The configuration is not empty
-            * The "count" option is present in the config
-            * If "node_options" is present, make sure its a dict of type `Dict[int, dict]`
+            * The `count" option is present in the config and is an integer
```
palango, line 84: should be **`** after **count**

scenario_player/utils/configuration/token.py
```diff
@@ -40,8 +40,9 @@ def __init__(self, loaded_definition: dict, token_info_fpath: pathlib.Path):
         self.dict = loaded_definition.get("token", {})
```
palango, line 40: I'd rename this from `dict`to `config`

scenario_player/utils/token.py
```diff
@@ -169,5 +169,4 @@ def save_token_configuration_to_file(token_file: str, token_data: TokenDetails)
 
 
     """
-    with open(token_file) as handler:
-        json.dump(token_data, handler)
+    token_file.write_text(json.dumps(token_data))
```
palango, line 172: Nice, didn't know that existed.

scenario_player/node_support.py
```diff
@@ -389,9 +397,108 @@ def _validate_options(self, options: Dict[str, Any]):
...
+                "Snapshots are only supported when token reuse is enabled "
+                "or a fixed token address is configured."
+            )
+
+    def _check_conditions(self):
+        all_nodes_stopped = all(not node_runner.is_running for node_runner in self._node_runners)
+        assert all_nodes_stopped, "Can't perform snapshot operations while nodes are running."
+        self.check_scenario_config()
+
+    def _get_snapshot_dirs(self, ensure: bool = False) -> List[Path]:
```
palango, line 431: Is this used with `ensure = True` anywhere?

scenario_player/tasks/execution.py
ulope: It may be useful to limit the concurrency in some cases (for example: when transferring via a hub only do up to X in parallel...)

scenario_player/utils/configuration/token.py
ulope: Yeah, all those config classes are a mess of several re-re-re-factorings by multiple people piling up on top of each other...

scenario_player/utils/token.py
ulope: Pathlib is one of my favorite improvements from Python 3.

scenario_player/node_support.py
ulope: Nope, good catch.

christianbrb:
> 
> 
> Looks good, some additional docs would be nice. especially when it makes sense to use it, which kind of tasks interact well...

@ulope Is there documentation available now?

Issue #491, opened 2020-02-21
Reenable smoketests


palango:
In https://github.com/raiden-network/scenario-player/pull/489 it showed that the smoketest currently fail: https://app.circleci.com/jobs/github/raiden-network/scenario-player/3625

I disabled them to be able to land the PR quickly, but we need to re-enable the smoketests.

give a try to do the proper fix according to @konradkonrad like the smoketest in Raiden

timeboxed to 2 days

konradkonrad:
- [ ] Decide, if a local blockchain is sufficient, if yes
- [ ] Decide if a blockchain snapshot can help to speed up smoketest boot time

hackaugusto:
Action plan:

### Plan1

Use plain text password for the account key

##### Up side

Fast to implement

##### Down side

- Vandalism
- We rely on a public test chain for our CI, which may result in flakiness.  

### Plan 2

Reuse raiden smoketest ctx managers
- [ ] Refactor the `smoketest` command from `raiden.ui.cli`, moving the setup of the `synapse`, the blockchain, and the deployment of the smart contracts to a new context manager [this code](https://github.com/raiden-network/raiden/blob/afe6df3240fe4654636b7dccafb95a0de45d319a/raiden/ui/cli.py#L772-L811)
- [ ] Reuse this context manager in the `scenario_player.main`, adding a new `smoketest` command, and then
- [ ] Create basic functionality template in source, then use the CM `setup_raiden` output to and dump it to yaml..

    Benefit: This will smoketest the whole stack (including the parsing of the scenario configuration)
    Downside: Extra code for the template and maintance for the template

- [ ] Run the scenario
- [ ] ???
- [ ] Profit

##### Up side

This won't rely on the internet, it is a self contained local test.
Doens't need to manage the private key of a test net in the CI server

##### Down side

A bit more of code / time / coupling among Raiden and the SP

fredo:
In [#5967](https://github.com/raiden-network/raiden/pull/5967) a job was added to Raiden CI to download the SP and cross test unit tests and smoketest. 

If you add a make rule `smoketest` then there is already commented code which just needs to be uncommented once the make rule is filled with life. 
https://github.com/fredo/raiden/blob/62fe02b56c0416152267ad94e3d96c3489471fda/.circleci/config.yml#L373-L377

PR #508, opened 2020-03-02
Re-enable smoketest

konradkonrad:
Part of #491

The account from `.circleci/smoketest.keystore` is funded and smoketests can be re-enabled.


konradkonrad:
Note: this actually _only_ works for PRs/branches that are pushed to `raiden-network/scenario-player` -- forked PRs don't have access to the environment variable `SMOKETEST_PW`.

Issue #490, opened 2020-02-20
Reenable skipped tests


palango:
- [ ] List here which tests are skipped and why, see [here](https://app.circleci.com/jobs/github/raiden-network/scenario-player/3633)
- [ ] Un-skip and fix it
- [ ] In case you cannot un-skip a certain list here and find someone to discuss to delete it

Work on it after #502 and #497 
Example: 
We need to reenable them:
- `TestNewTransactionEndpoint.test_new_transaction_calls_correct_rpc_client_function`
- `TestNewTransactionEndpoint.test_endpoint_requires_parameter_and_expects_types[None missing - correct types]`

Dominik1999:
start after #495 is completed

PR #511, opened 2020-03-03
Simplified code: Removed transactions server.

### LINKED ISSUE #502: Remove `/rpc/transactions` and `/rpc/contract` services
In https://github.com/raiden-network/scenario-player/issues/495 we decided that we don't need the service architecture at the moment and want to simplify the code. This tasks tracks the implementation.

- [ ] Remove the `/rpc/transactions` endpoint
- [ ] Remove the `/rpc/contract` endpoint
- [ ] Remove service architecture infrastructure

hackaugusto:
This is a big PR that simplifies quite a lot of the code. Currently a WIP. Fixes #502

The  transaction server is just adding complexity to the code base
without a clear benefit to the system. This removes the existing
server to manage transactions.

- Streamlined the scenario player code, making dependencies of setup
  transactions clearer and straigh forward to understand.
- Introduces some small optimizations which allow for deployment and
  minting across the utility and network tokens.
- uses the proxies from the Raiden code base which properly detect
  errors, wait for transactions confirmations.
- Improved the typing of the functions and moved small methods to plain
  functions.
- Removed the extra abstraction for Contracts
- Remove the services code, their tests, manager, and corresponding
  exceptions.
- Increase the limits of minting / approving for the scenario's player
  account to a large number, this will benefit multiple sequential runs
  using the same account. (Previously an arbitrary value of twice the
  current requirement was used, which would help only for the next
  scenario player run *if* using the same scenario file).

Closes https://github.com/raiden-network/scenario-player/issues/490


scenario_player/main.py
```diff
@@ -338,22 +335,6 @@ def __exit__(self, type, value, traceback):
                 self.ui_greenlet.join()
 
 
-class ServiceProcessManager:
```
hackaugusto: This is not necessary anymore since there is no `ServiceProcess` to be managed.

scenario_player/runner.py
```diff
@@ -2,42 +2,172 @@
...
 
 if TYPE_CHECKING:
     from scenario_player.tasks.base import Task, TaskState
 
 log = structlog.get_logger(__name__)
 
 
+def get_udc_and_corresponding_token(
+    udc_settings: UDCSettingsConfig, settings: SettingsConfig, proxy_manager: ProxyManager
+) -> Tuple[UserDeposit, Token]:
```
hackaugusto: This corresponds to the old [`get_udc_and_token`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/utils/legacy.py#L301-L326), however now it returns the proxies from Raiden instead of yet another proxy.

The type signature of the function was slightly simplified by always requiring the `assert udc_settings.enable`, which means the return type does not need `Optional` anymore.

scenario_player/runner.py
```diff
@@ -2,42 +2,172 @@
...
+        UserDepositAddress(to_bytes(udc_address)), "latest"
+    )
+
+    token_address = user_deposit_proxy.token_address("latest")
+    token_proxy = proxy_manager.token(token_address, "latest")
+
+    return user_deposit_proxy, token_proxy
+
+
+def determine_run_number(scenario_dir: Path) -> int:
```
hackaugusto: This corresponds to the [`ScenarioRunner.determine_run_number`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L121-L135), this doesn't need to be a method since it does not required internal state. Moved it out and simplified the signature/code.

scenario_player/runner.py
```diff
@@ -2,42 +2,172 @@
...
+        run_number = int(run_number_file.read_text()) + 1
+    else:
+        run_number = 0
+
+    run_number_file.write_text(str(run_number))
+
+    return run_number
+
+
+def make_session(auth: str, settings: SettingsConfig) -> Session:
```
hackaugusto: This corresponds to a small setup done in the [`ScenarioRunner.__init__`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L101-L105). It doesn't need state from `ScenarioRunner` so it does not need to be in there. Moving it out makes it easier to read `__init__`.

scenario_player/runner.py
```diff
@@ -2,42 +2,172 @@
...
+    session = Session()
+    if auth:
+        session.auth = cast(Tuple[str, str], tuple(auth.split(":")))
+    session.mount("http", TimeOutHTTPAdapter(timeout=settings.timeout))
+    session.mount("https", TimeOutHTTPAdapter(timeout=settings.timeout))
+
+    return session
+
+
+def wait_for_token_network_discovery(
```
hackaugusto: This corresponds to the [`ScenarioRunner.wait_for_token_network_discovery`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L180-L226) which again doesn't need the instance state.

scenario_player/runner.py
```diff
@@ -49,12 +179,11 @@ def __init__(
         task_state_callback: Optional[
             Callable[["ScenarioRunner", "Task", "TaskState"], None]
         ] = None,
-    ):
-        from scenario_player.node_support import RaidenReleaseKeeper, NodeController
```
hackaugusto: Moved to top-level.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
             seed = seed_file.read_text().strip()
         return seed
 
-    def select_chain(self, chain_urls: Dict[str, List[str]]) -> Tuple[str, List[str]]:
-        """Select a chain and return its name and RPC URL.
+    def ensure_token_network_discovery(self) -> ChecksumAddress:
```
hackaugusto: Corresponds to the old [`ensure_token_network_discovery`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L228-L235). Improve it by making sure the nodes don't diverge (all report the token network address).

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
             )
 
-    def wait_for_token_network_discovery(self, node) -> ChecksumAddress:
-        """Check for token network discovery with the given `node`.
+        return token_network_addresses.pop()
 
-        By default exit the wait if the token has not been discovered after `n` seconds,
-        where `n` is the value of :attr:`.timeout`.
+    def run_scenario(self):
+        self.node_controller.initialize_nodes()
```
hackaugusto: The node controller initialization used to be spread out across different threads [here](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L381-L383) and [here](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L296-L297). This places the initialization in the same thread as the test and groups the calls to the controller.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
+                f"values : {token_network_addresses}"
             )
 
-    def wait_for_token_network_discovery(self, node) -> ChecksumAddress:
-        """Check for token network discovery with the given `node`.
+        return token_network_addresses.pop()
 
-        By default exit the wait if the token has not been discovered after `n` seconds,
-        where `n` is the value of :attr:`.timeout`.
+    def run_scenario(self):
```
hackaugusto: This corresponds to the old [`ScenarioRunner.run_scenario`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L291-L306)

It was simplified by removing the unused variables.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
+        # Wait for either a crash in one of the Raiden nodes or for the
+        # scenario to exit (successfully or not).
+        gevent.joinall({node_monitor, scenario}, raise_error=True, count=1)
 
-            except HTTPError as e:
-                # We explicitly handle 404 Not Found responses only - anything else is none
-                # of our business.
-                if e.response.status_code != 404:
-                    raise
+    def setup_environment_and_run_main_task(self, node_addresses, node_starter) -> None:
```
hackaugusto: This corresponds to the old [`ScenarioRunnner._setup`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L237-L289). Renamed because in the end o f the setup function the scenario was executed.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
-                log.error("Couldn't register token with network", code=code, message=msg)
-                raise TokenRegistrationError(msg)
-
-        self.token_network_address = self.ensure_token_network_discovery()
-        log.info(
-            "Token Network Discovery Completed", token_network_address=self.token_network_address
-        )
-        log.info("Setup done")
-        # Start root task
-        root_task_greenlet = gevent.spawn(self.root_task)
```
hackaugusto: This extra greenlet is unecessary, calling `self.root_task` has the same effect.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
-        root_task_greenlet.name = "root_task"
-        greenlets.add(root_task_greenlet)
-        try:
-            gevent.joinall(greenlets, raise_error=True)
-        except BaseException as exc:
-            log.exception("Greenlet Error", exception=exc)
-            if not root_task_greenlet.dead:
-                # Make sure we kill the tasks if a node dies
-                root_task_greenlet.kill()
```
hackaugusto: This only makes sure the `root_task_greenlet` dies if the [`monitor thread dies`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L297). This should be here but at `run_scenario` to keep spacial locality of the threads.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
-            raise
+        return ether_transfer_transactions
 
-    def run_scenario(self):
-        mint_gas = GAS_LIMIT_FOR_TOKEN_CONTRACT_CALL * 2
+    def setup_mint_user_deposit_tokens_for_distribution(
```
hackaugusto: This used to be the last part of [`_initialize_udc`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L357-L371). The method was split so that concerns are not mixed, it simplifies the signature and makes it easier to understand by reading the method name. Now only needs to return the set of transaction hashes.

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
+            to_checksum_address(self.client.address),
+            minimum_balance=required_allowance,
+            maximum_balance=UINT256_MAX,
+        )
+        if mint_tx:
+            orchestration_user_deposit_token_transactions.add(mint_tx)
+
+        return orchestration_user_deposit_token_transactions
+
+    def setup_raiden_token_balances(
```
hackaugusto: This used to be in the middle of [`_initialize_nodes`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L382-L405), now it is on its own method which allows for concurrent execution, and a clearer interface. (The rest of the code was managing nodes which has been grouped together).

scenario_player/runner.py
```diff
@@ -153,279 +256,260 @@ def local_seed(self) -> str:
...
-        return mint_tx
+        return transaction_hashes
 
-    def _initialize_udc(
-        self, gas_limit: int, node_count: int
-    ) -> Tuple[Set[TransactionHash], Optional[Contract], bool]:
-        our_address = to_checksum_address(self.client.address)
-        udc_settings = self.definition.settings.services.udc
-        udc_enabled = udc_settings.enable
+    def setup_raiden_nodes_with_sufficient_user_deposit_balances(
```
hackaugusto: This used to be part of the [`_initialize_scenario_token`](https://github.com/raiden-network/scenario-player/blob/f208ed6cb4f713c1efc921b10458365bcf918b5d/scenario_player/runner.py#L325-L330), which did minting for both the user deposity utility token and for the token network token. Now split.

scenario_player/utils/token.py
```diff
@@ -15,456 +17,249 @@
...
-
-        log.info(f"Requesting '{action}' call", **payload)
-        resp = self.interface.post(f"spaas://rpc/contract/{action}", json=payload)
-        resp.raise_for_status()
-        resp_data = resp.json()
-        tx_hash = resp_data["tx_hash"]
-        log.info(f"'{action}' call succeeded", tx_hash=tx_hash)
-        return tx_hash
-
-    def mint(
```
hackaugusto: this method is condition, it was moved to a plain function and renamed to `token_maybe_mint`

scenario_player/utils/token.py
```diff
@@ -15,456 +17,249 @@
...
+def token_maybe_mint(
+    token_proxy: Token, target_address, minimum_balance: int, maximum_balance: int
+) -> Optional[TransactionHash]:
+    current_balance = token_proxy.balance(target_address)
 
+    if minimum_balance > current_balance:
+        mint_amount = maximum_balance - current_balance
+        return token_proxy.mint(amount=mint_amount, target_address=target_address)
 
-class Token(Contract):
```
hackaugusto: All of this code was duplicated and removed in favor of `raiden.network.proxies.token.Token`

scenario_player/utils/token.py
```diff
@@ -15,456 +17,249 @@
...
+    user_deposit_proxy: UserDeposit,
+    target_address: ChecksumAddress,
+    minimum_effective_deposit: int,
+    maximum_funding: int,
+) -> Optional[TransactionHash]:
+    """Make a deposit at the given `target_address`.
+
+    The amount of tokens depends on the scenario definition's settings.
 
-    def update_allowance(self) -> Tuple[Optional[TransactionHash], int]:
```
hackaugusto: this code was condition, moved out of the method into a plain function named `userdeposit_maybe_increase_allowance`

scenario_player/utils/token.py
```diff
@@ -15,456 +17,249 @@
...
+    The file contents should be at least::
+
+        {
+            "name": "<token name>",
+            "address":, "<contract address>",
+            "block": <deployment block},
         }
-        return self.transact("allowance", params), required_allowance
-
-    def deposit(self, target_address) -> Union[TransactionHash, None]:
```
hackaugusto: again, conditional function, now named `userdeposit_maybe_deposit`

scenario_player/runner.py
```diff
@@ -2,46 +2,259 @@
...
 
 
+# The `mint` function checks for overflow of the total supply. Here we
+# have a large enough number for both scenario runs and the scenario
+# orchestration account capacity.
+NUMBER_OF_RUNS_BEFORE_OVERFLOW = 2 ** 64
+ORCHESTRATION_MAXIMUM_BALANCE = UINT256_MAX // NUMBER_OF_RUNS_BEFORE_OVERFLOW
+
+
+def is_udc_enabled(udc_settings: UDCSettingsConfig):
```
hackaugusto: This predicate corresponds to this expression:

https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/runner.py#L346

scenario_player/runner.py
```diff
@@ -2,46 +2,259 @@
...
+    should_deposit_ud_token = udc_settings.token.deposit
+    return udc_settings.enable and should_deposit_ud_token
+
+
+def wait_for_nodes_to_be_ready(node_runners, session):
```
hackaugusto: I just moved this loop out, since we don't really need it in a separate greenlet:

https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/runner.py#L403-L416

scenario_player/runner.py
```diff
@@ -2,46 +2,259 @@
...
+            while True:
+                try:
+                    if session.get(url).ok:
+                        break
+                except requests.exceptions.RequestException:
+                    pass
+                gevent.sleep(0.5)
+
+
+def get_udc_and_corresponding_token_from_dependencies(
```
hackaugusto: This corresponds to this code:

https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/utils/legacy.py#L213-L240

scenario_player/runner.py
```diff
@@ -2,46 +2,259 @@
...
+        UserDepositAddress(to_canonical_address(udc_address)), "latest"
+    )
+
+    token_address = userdeposit_proxy.token_address("latest")
+    user_token_proxy = proxy_manager.custom_token(token_address, "latest")
+
+    return userdeposit_proxy, user_token_proxy
+
+
+def get_token_network_registry_from_dependencies(
```
hackaugusto: This sort of replaces this RPC call:

https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/runner.py#L257-L271

There is a problem here, since the scenario player version of the contracts may be different than the version used by the raiden client. This is also true for the user deposit and token smart contracts.

We decided that instead of having a mix of solutions that work sometimes and not other we have decided to implement the simple solution first, which is to just rely on manual synchronization of the dependencies. This should not be a big problem since package releases are spaced out.

A proper solution would require a new command in raiden to expose the node's configuration.

scenario_player/runner.py
```diff
@@ -2,46 +2,259 @@
...
+                raise TypeError(f"Unexpected response type from API: {data!r}")
+
+            return ChecksumAddress(data)
+
+    # We could not assert that our token network was registered within an
+    # acceptable time frame.
+    raise TokenNetworkDiscoveryTimeout
+
+
+def maybe_create_token_network(
```
hackaugusto: This corresponds to https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/runner.py#L264

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
             seed = seed_file.read_text().strip()
         return seed
 
-    def select_chain(self, chain_urls: Dict[str, List[str]]) -> Tuple[str, List[str]]:
-        """Select a chain and return its name and RPC URL.
+    def ensure_token_network_discovery(
```
hackaugusto: Same as https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/runner.py#L228 with an small improvement, now it not only checks there is a token network, but also that all have the same one

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
...
+            if to_canonical_address(address) != Address(token_network_addresses):
+                raise RuntimeError(
+                    f"Nodes diverged on the token network address, there should be "
+                    f"exactly one token network available for all nodes. Current "
+                    f"values : {to_hex(token_network_addresses)}"
+                )
+
+    def run_scenario(self) -> None:
+        with Janitor() as nursery:
+            self.node_controller = NodeController(self, self.definition.nodes, nursery)
```
ulope: A bit nitpicky: I'm not a fan of creating instance variables outside of the call scope of `__init__`. That's just asking for `AttributeError`s later on.

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
...
+        root_task()
 
-            setup = nursery.spawn_under_watch(
-                self._setup, mint_gas, node_count, node_addresses, fund_tx, node_starter
+    def setup_raiden_nodes_ether_balances(
+        self, pool: Pool, node_addresses: Set[ChecksumAddress]
+    ) -> Set[Greenlet]:
+        """ Makes sure every Raiden node has at least `NODE_ACCOUNT_BALANCE_MIN`. """
+
+        greenlets: Set[Greenlet] = set()
```
ulope: Is the set necessary? 
Shouldn't the pool keep track of the added greenlets by itself?

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
...
+        mint_greenlet = pool.spawn(
+            token_maybe_mint,
+            token_proxy,
+            to_checksum_address(self.client.address),
+            minimum_balance=required_allowance,
+            maximum_balance=ORCHESTRATION_MAXIMUM_BALANCE,
+        )
 
-        return mint_tx
+        return {allowance_greenlet, mint_greenlet}
```
ulope: Same question as above

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
...
-                    "startgas": 21_000,
-                }
-                resp = self.service_session.post("spaas://rpc/transactions", json=params)
-                tx_hash = resp.json()["tx_hash"]
-                fund_tx.add(tx_hash)
+        token_definition = self.definition.token
+        reuse_token_from_file = token_definition.reuse_token
+        token_address = to_canonical_address(token_definition.address)
+        deploy_new = not (reuse_token_from_file or token_address)
+        token_data_path = str(self.data_path.joinpath("token.info"))
```
ulope: Why not keep this a Path instance?

scenario_player/runner.py
```diff
@@ -153,289 +335,326 @@ def local_seed(self) -> str:
...
-                    url = f"http://{node_runner.base_url}/api/v1/address"
-                    while True:
-                        try:
-                            if self.session.get(url).ok:
-                                break
-                        except requests.exceptions.RequestException:
-                            pass
-                        gevent.sleep(0.5)
-            log.info("All nodes ready")
+            if reuse_token_from_file:
```
ulope: This can never be true unless I'm reading the code wrong.

scenario_player/utils/token.py
```diff
@@ -1,483 +1,182 @@
...
-        self.contract_data = {
-            "token_contract": address,
-            "name": contract_info.get("name") or contract_name,
-        }
-        self.contract_proxy = self._local_rpc_client.new_contract_proxy(
-            contract_info["abi"], address
+    if balance < minimum_balance:
+        eth_transfer = EthTransfer(
+            to_address=target,
+            value=minimum_balance - balance,
```
ulope: This used to fund to `NODE_ACCOUNT_BALANCE_FUND` so we don't have to re-fund on every run in the reuse case.

scenario_player/runner.py
hackaugusto: This one is necessary, this thread mints tokens for the orchestration account, which is then transferred by `setup_raiden_nodes_with_sufficient_user_deposit_balances`. These greenlets are used for synchronization since the token transfer can only be performed after the mint transaction is

scenario_player/runner.py
hackaugusto: It turned out to be less code with the same functionality.

scenario_player/runner.py
hackaugusto: humm, it should be the same as this:

https://github.com/raiden-network/scenario-player/blob/edd041e816776c141a1182f089ab6bb75a976e91/scenario_player/utils/token.py#L265-L269

no?

hackaugusto: I guess I can change the code to something like:
```python
if token_address:
  # instantiate proxy
elif reuse_token and token_info.exists():
  # load json
else:
  token = deploy()
  if reuse: save_json()
```

scenario_player/utils/token.py
hackaugusto: That didn't change:

https://github.com/raiden-network/scenario-player/blob/e363a5bc9663d23d65ce0adc84f24525d9868771/scenario_player/runner.py#L500

hackaugusto: Oh, wait, these are different constants.

scenario_player/runner.py
hackaugusto: Agreed, I don't like partially initialized objects. I did not try to fix that though, but IMO this should go in a follow up.

scenario_player/runner.py
ulope: Hm ok, not important.

Issue #468, opened 2020-01-14
Reuse the Janitor from Raiden - replace mirakuru


hackaugusto:
The current implementation is using a hot-loop to monitor the state of spawned Raiden processes.

https://github.com/raiden-network/scenario-player/blob/511a33ab6fbb234182184761ee44d1bf4e6912db/scenario_player/node_support.py#L547-L562

This is unnecessary since the gevent library provides an `AsyncResult` for the exit status of a `Popen` subprocess.

I'd recommend just reusing Raiden's Janitor implementation (instead of mirakuru and the current existing monitor), which has less code and implements the same functionality while handling corner cases (namely cleaning resources on exit):

https://github.com/raiden-network/raiden/blob/919115d4f140403e3b12e57bdf93bfcd80e7b5ff/tools/debugging/stress_test_transfers.py#L138-L218

Implementation:
- [x] Decide whether we want to check free ports during startup?
- [ ] Do we need healthchecks for running processes? (Probably not necessary according to @karlb and @palango) 
- [x] we do need a startup check, wait for nodes being up before starting the tasks
- [x] Replace mirakuru with Janitor

Testing:
- [ ] Run scenario player smoketest
- [x] Run simple scenario to test basic startup
- [x] Run scenario that stops and starts raiden node as part of the scenario
- [x] Add bug to Raiden and run scenario to test error handling. Expected:
  - [x] Original exception must be visible
  - [x] Other processes must be stopped cleanly
  - [x] No raiden processes must be left running when SP has finished
- [x] Processes are stopped when leaving the UI before the scenario has finished

palango:
So after looking into this a bit I'm torn if this is worth implementing.

The idea was to get rid of mirakuru all together.
We currently use it in three places:
- In the SP to run and monitor Raiden nodes
- In the integration tests to run Ethereum nodes
- In the integration tests to run Matrix servers

We have two slightly different implementations in raiden and the SP, and it makes sense to unify them.

However, in order to get rid of mirakuru we'd basically reimplement the whole thing with some changes to process handling, which I don't think is useful.

PR #514, opened 2020-03-09
Use Raiden Janitor instead of mirakuru

karlb:
Implements https://github.com/raiden-network/scenario-player/issues/468 in the most basic way. Followups to be discussed (feel free to add items during review):
- [ ] readd health check
- [ ] send SIGKILL to nodes that won't stop by themselves (the SIGKILL in https://github.com/raiden-network/raiden/pull/5974 might be enough)

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Tuple, cast
 
 import gevent
+import requests
 import structlog
 from eth_typing import ChecksumAddress
 from eth_utils import encode_hex, is_checksum_address, to_canonical_address, to_checksum_address
@@ -399,9 +400,16 @@ class ScenarioRunner:
             raise
 
         def wait():
-            for node_runner in self.node_controller._node_runners:
-                while not node_runner.executor.after_start_check():
-                    gevent.sleep(0.2)
+            with gevent.Timeout(100):
+                for node_runner in self.node_controller._node_runners:
+                    url = f"http://{node_runner.base_url}/api/v1/address"
+                    while True:
+                        try:
+                            if self.session.get(url).ok:
+                                break
+                        except requests.exceptions.RequestException:
+                            pass
+                        gevent.sleep(0.5)
             log.info("All nodes ready")
 
         node_starter = nursery.spawn_under_watch(wait)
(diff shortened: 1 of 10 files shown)


scenario_player/runner.py
```diff
@@ -371,7 +359,7 @@ def _initialize_udc(
         return ud_token_tx, udc_ctr, should_deposit_ud_token
 
     def _initialize_nodes(
-        self,
+        self, nursery: Any
```
palango: Can't we type this properly?

scenario_player/runner.py
```diff
@@ -404,7 +392,28 @@ def _initialize_nodes(
...
-        node_starter = self.node_controller.start(wait=False)
+        try:
+            for node_runner in self.node_controller._node_runners:
+                node_runner.start()
+        except Exception:
+            log.error("failed to start", exc_info=True)
+            raise
+
+        def wait():
+            with gevent.Timeout(100):
```
palango: Maybe pull this into a constant for more visibility (I guess right now we're fast enough, but we could run into this)

scenario_player/runner.py
```diff
@@ -276,34 +275,23 @@ def _setup(self, mint_gas, node_count, node_addresses, fund_tx, node_starter, gr
...
-        greenlets.add(root_task_greenlet)
-        try:
-            gevent.joinall(greenlets, raise_error=True)
-        except BaseException as exc:
-            log.exception("Greenlet Error", exception=exc)
-            if not root_task_greenlet.dead:
-                # Make sure we kill the tasks if a node dies
-                root_task_greenlet.kill()
-            raise
+        self.root_task()
```
palango: Why don't we need to start this in a greenlet any more?

scenario_player/runner.py
karlb: The `Nursery` is defined inside the `Janitor` as a closure. I don't think mypy can access that. We can probably pull that out into a separate class with some additional constructor args. But that has to be done on the Janitor side first.

scenario_player/runner.py
karlb: This whole function is already run in a greenlet. I didn't see a need to add an additional greenlet. We could probably have removed that greenlet even without the other changes.

scenario_player/runner.py
palango: There's `class Nursery(ABC):` for that reason I think.

karlb: Good point, I forgot about that!

PR #521, opened 2020-03-10
Don't catch GreenletExit

karlb:
Otherwise we can't kill the tasks when we want to exit.

Also catch `Exception` instead of `BaseException`, since this is usually
the right thing to do and there is no reason given for catching
`BaseException`.

### DIFF
--- a/scenario_player/tasks/base.py
+++ b/scenario_player/tasks/base.py
@@ -40,11 +40,7 @@ class Task:
     _name: str
 
     def __init__(
-        self,
-        runner: scenario_runner.ScenarioRunner,
-        config: Any,
-        parent: "Task" = None,
-        abort_on_fail=True,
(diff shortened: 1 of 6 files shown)


karlb:
This has some impact success status. Don't review/merge, yet.

scenario_player/tasks/base.py
```diff
@@ -70,9 +70,7 @@ def __call__(self, *args, **kwargs):
         self._start_time = time.monotonic()
         try:
             return self._run(*args, **kwargs)
-        except gevent.GreenletExit:
-            pass
-        except BaseException as ex:
```
ulope: The thinking was that any exception (including `KeyboardInterrupt`) should terminate a task. That could be fixed by adding a signal handler as we have in Raiden.

karlb: I went back to `BaseException` but made sure that it's reraised after the task is set to failed.

PR #523, opened 2020-03-11
Remove dead code

hackaugusto:
I ran `vulture` in the code, these are the pieces that are not used.


hackaugusto:
I tested it with bf1

PR #530, opened 2020-03-16
Upgrade matrix-synapse dev dependency

konradkonrad:
This upgrades the `matrix-synapse` dependency to be in sync with `raiden-network/raiden`. Merge this only after https://github.com/raiden-network/raiden/pull/5983


palango:
Incorporated into https://github.com/raiden-network/scenario-player/pull/534

PR #531, opened 2020-03-18
removed curses ui

hackaugusto:


ulope:
(Caveat: I'm obviously a bit biased since I wrote most of the code)

For me the question is why remove it? Does removing it make something so much better that it outweighs the loss in functionality?
My main reasons for not removing it are:
- The UI code isn't a big maintenance burden. AFAIK it specifically has rarely if ever been the cause of any problems or bugs in the SP.
- The UI is very useful esp. while writing scenarios to see what is going on (e.g. with more complex scenarios that use nested task groups with repeat etc.) Only having the logs makes it very difficult to visualize where in a scenario one currently is.

But I guess it would be most useful to ask the people from the testing team that have written most of the scenarios. 
I know @czepluch has done a lot of that. Not sure who else.

czepluch:
As @ulope mentions it's been quite useful while initially writing the scenarios. I think especially if someone who has not been working a lot with the scenarios were to start writing scenarios it would be hard to get started without the curses ui. A second thing that is nice is that it's easy to see some measurements regarding transfer time etc.
I do however think that after some getting used to, the logs are just fine. 
I have no strong opinion, but I agree with Ulo, that it seems unnecessary to remove it.

PR #532, opened 2020-03-18
Removed ConfigMapping

hackaugusto:
Changes:

- Simplifies the code by removing the unecessary abstraction
  `ConfigMapping` and the custom assert method.
- Fixes a few bugs, were a default value for missing key was not set,
  which could lead to `KeyError`s. (Failling early could have been the
  better option, but for now this only makes the code consistent).
- Config objects don't inherit from ConfigMapping, so instead of having
  three ways of accessing the data, now it should only use the
  properties. (Previously `self.dict[attr]`, `self[attr]`, and `self.attr` were
  used).


scenario_player/utils/configuration/settings.py
```diff
@@ -89,23 +87,25 @@ def deposit(self) -> bool:
 
         Defaults to True.
         """
-        return self.get("deposit", True)
+        flag = self.dict.get("deposit", True)
+        assert isinstance(flag, bool)
+        return flag
```
karlb: I'm looking forward to turning some of these properties into plain attributes. E.g. the `deposit` attribute could probably just be a
```python
self.deposit = loaded_definition.get("deposit", True)
```
line in `__init__`.

hackaugusto: I'm doing that for a follow up already

PR #536, opened 2020-03-30
Environments for scenarios

karlb:
The scenarios should be runnable in different environments, not just in our development setup. To make that possible, this PR reads and environment description file and uses the values to template the scenario descriptions.

Do not merge this PR, today (Monday). I have to make sure that the [scenarios](https://github.com/raiden-network/raiden/pull/6033) and CI setup match this, before it can be used for the nightly run.

See https://github.com/raiden-network/team/issues/682


environment/development.json
```diff
@@ -0,0 +1,7 @@
+{
+	"environment_type": "development",
+	"transport_servers": ["https://transport01.raiden.network", "https://transport02.raiden.network", "https://transport03.raiden.network", "https://transport04.raiden.network"],
+	"pfs_with_fee": "https://pfs-goerli-with-fee.services-dev.raiden.network",
+	"pfs_without_fee": "https://pfs-goerli.services-dev.raiden.network",
+	"eth_rpc_endpoint": "http://parity.goerli.ethnodes.brainbot.com:8545"
```
palango: Not for this PR, but for the future: Would be nice to make this a list as well and let the client chose randomly, so we would test with geth as well.

scenario_player/main.py
```diff
@@ -195,28 +204,39 @@ def run(
...
-        data_path,
-        auth,
-        password,
-        keystore_file,
-        scenario_file,
-        notify_tasks,
-        enable_ui,
-        password_file,
-        log_file_name,
+        chain=chain,
```
palango: Nice that you do these small improvements!

PR #540, opened 2020-04-02
`claim` now takes environment file, like `run`

karlb:
For claiming, we only need the eth-rpc-endpoint from the environment
file. This is the reason why I previously used and `--eth-rpc-endpoint`
option, like Raiden does. But I realized that consistency with the `run`
command is more useful than being minimal and consistent with Raiden in
this case.

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -144,21 +144,20 @@ def data_path_option(func):
     return wrapper
 
 
-def eth_rpc_option(func):
-    """Decorator for adding '--eth-rpc-endpoint' to subcommands."""
+def environment_option(func):
+    """Decorator for adding '--environment' to subcommands."""
 
     @click.option(
-        "--eth-rpc-endpoint",
-        multiple=False,
-        required=True,
-        help=(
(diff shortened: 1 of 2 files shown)


scenario_player/main.py
```diff
@@ -144,21 +144,20 @@ def wrapper(*args, **kwargs):
...
+        help="A JSON file containing the settings for Eth-RPC, PFS, "
+        "transport servers, env-type...",
+        show_default=True,
+        type=click.File(),
     )
     @functools.wraps(func)
-    def wrapper(*args, **kwargs):
-        return func(*args, **kwargs)
+    def wrapper(*args, environment: IO, **kwargs):
+        return func(*args, environment=_load_environment(environment), **kwargs)
```
ulope, line 160: FYI: This could be implemented as a click `ParamType` subclass. But this should be fine as well.

Example in Raiden: https://github.com/raiden-network/raiden/blob/07b21ac2bf8a935e26347267b2379e63726e630b/raiden/utils/cli.py#L225-L232

PR #542, opened 2020-04-16
Avoid inconsistent RECLAIM_MIN_BALANCE

karlb:
Before this commit, `RECLAIM_MIN_BALANCE` was not the same as the tx
cost for the reclaim. If it's bigger, we would not reclaim some small
funds. If it's smaller, we would get an error because we would try to
reclaim a negative value.

The obvious fix is to use the tx cost as the minimum reclaim balance.

### DIFF
--- a/scenario_player/utils/legacy.py
+++ b/scenario_player/utils/legacy.py
@@ -22,7 +22,6 @@ from raiden.network.rpc.client import EthTransfer, JSONRPCClient, TransactionSen
 from raiden.utils.typing import ChecksumAddress, PrivateKey
 from scenario_player.exceptions import ScenarioTxError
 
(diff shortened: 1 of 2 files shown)


scenario_player/utils/legacy.py
```diff
@@ -149,15 +148,16 @@ def reclaim_eth(
 
     txs = []
     reclaim_amount = 0
+    reclaim_tx_cost = web3.eth.gasPrice * VALUE_TX_GAS_COST
```
hackaugusto, line 151: `web3.eth.gasPrice` is a property that hides a RPC call ... that means this value may change just a few lines below when the transaction is actually sent. Could you please reuse the same value to avoid incoherences? (the gas price could go slightly up and the cost of the transaction too, so the transaction would fail even the check is in-place)

Otherwise LGMT

PR #551, opened 2020-05-05
Withdraw tokens from channels during ERC20 reclaim

### LINKED ISSUE #547: Support withdrawing from channels created by nodes
To properly reclaim all tokens invested into a scenario, all channels have to be closed and settled before reclaiming the tokens. Settling channels can also be helpful to retrieve tokens in other testing cases. Implement this for the SP now, but try to keep requirements for the code low, so that we can repurpose it later.

Not reading the nodes db makes this work in cases with lost, damaged or outdated dbs. The downside is that empty balance proofs will be used to close and settle. This is not a problem in our test setups, but will limit the general usability of this code.

Alternatively, we can use withdrawing instead of closing and settling to reduce the amount of transactions and the waiting time.

- [x] Take list of keys and a TokenNetwork address as input
- [x] Fetch events from TokenNetwork to find open channels
- ~~Close all channels~~
- ~~Wait for settlement window to pass~~
- ~~Settle all channels~~
- [x] Withdraw tokens

karlb:
Thanks @hackaugusto for the awesome idea of using withdraw instead of close/settle.

When using geth, fetching the events times out. It works fine with parity and infura.

Closes #547


scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,425 @@
...
+    web3: Web3,
+    token_network_address: TokenNetworkAddress,
+    deploy: DeployedContracts,
+) -> List[dict]:
+    """ Read ChannelOpened events to get the channel_ids and participants """
+    new_channel_abi = contract_manager.get_event_abi(CONTRACT_TOKEN_NETWORK, ChannelEvent.OPENED)
+    topics = [
+        encode_hex(event_abi_to_log_topic(new_channel_abi)),  # type: ignore
+        None,
+        encode_hex(bytes([0] * 12) + to_canonical_address(candidate.address)),
```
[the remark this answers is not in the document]
hackaugusto: The topics are not set correctly, there is not guarantee that `candidate.address` will be the third fields instead of the second:

https://github.com/raiden-network/raiden-contracts/blob/d3becfb354e250979f1fd81b8966613eb8dd9ded/raiden_contracts/data/source/raiden/TokenNetwork.sol#L301-L306

This either requires two calls, one with the topics `[channel_open_abi, None, address]` and another for `[channel_open_abi, address, None]`, or the filter should just be  `[channel_open_abi]` and the events have to be filtered locally.  If you do filter locally, consider removing the filtering altogether, since this will be doing a linear search there is no guarantee this will be faster to process anyways, the only benefit it will have is reducing the response size.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,425 @@
...
+def _withdraw_all_from_channel(
+    candidate: ReclamationCandidate,
+    reclamation_candidates: List[ReclamationCandidate],
+    channel: dict,
+    token_network: TokenNetwork,
+):
+    # Check if channel still has deposits
+    details = token_network.detail_participants(
+        participant1=to_canonical_address(candidate.address),
+        participant2=to_canonical_address(channel["participant2"]),
```
hackaugusto, line 329: There is no guarantee that `participant2` is not the *candidate* address. First one has to make sure the partner address is used instead, something like:

```python
if channel['participant1'] == candidate.address:
  partner_address = channel['participant1']
elif channel['participant2'] == candidate.address:
  partner_address = channel['participant1']
else:
  raise RuntimeError('not a participant of the given channel')
```

the here you can use `partner_address` instead of `channel["participant2"]`.

hackaugusto: The same applies to all `channel["participant2"]` below.

scenario_player/main.py
```diff
@@ -466,10 +467,13 @@ def reclaim_eth(
     configure_logging_for_subcommand(construct_log_file_name("reclaim-eth", data_path))
 
     reclamation_candidates = get_reclamation_candidates(data_path, min_age)
+    address_to_candidate: Dict[Address, ReclamationCandidate] = {
+        to_canonical_address(c.address): c for c in reclamation_candidates
```
hackaugusto, line 471: I couldn't use a set of `ReclamationCandidate` because the class is not hashable, and I didn't want to go through the trouble of removing the dictionary instance now.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+    deploy = get_contracts_deployment_info(chain_id, RAIDEN_CONTRACT_VERSION)
+    assert deploy
+
+    address_to_proxy_manager: Dict[ChecksumAddress, ProxyManager] = dict()
+    planned_withdraws: Dict[ChecksumAddress, Tuple[BlockNumber, TokenAmount]] = dict()
+
+    log.info("Checking chain for deposits in UserDeposit contact")
+    for node in reclamation_candidates:
+        if node.address not in address_to_proxy_manager:
+            client = JSONRPCClient(web3, node.privkey)
```
hackaugusto, line 136: Hummm ... I didn't see these before, I think this code was just moved instead of added to the PR so I didn't review it before. This has to be fixed otherwise we will most likely have `nonce` problems, only one instance of a `JSONRPCClient` can exist at any time because there is no external coordination.

Even if this instances don't leave the function this should be fixed.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+            address_to_proxy_manager[node.address] = get_proxy_manager(client, deploy)
```
hackaugusto, line 137: Same thing with the proxy manager, the locks are effect less if there are more than once instance of each proxy.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+    account: Account,
+    eth_rpc_endpoint: URI,
+):
+    web3 = Web3(HTTPProvider(eth_rpc_endpoint))
+
+    reclaim_amount = 0
+
+    log.info("Checking chain for claimable tokens", token_address=token_address)
+    for node in reclamation_candidates:
+        client = JSONRPCClient(web3, node.privkey)
```
hackaugusto, line 208: Same problem here ... This also has an unnecessary runtime cost to the Ethereum node, since each of these classes perform runtime checks which require REST calls, which may or not be expensive. But as we have experience before, the nodes are not able to handle a lot of concurrent action (specially with logs), so we should avoid these unecessary extra calls.

Additionally, it looks like these objects are being used through out the functions, so it makes sense to have a single utility to instantiate them all.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+                    )
+                )
+            )
+        (node.node_dir / "reclaimed").touch()
+
+    wait_for_txs(web3, txs, 1000)
+    log.info("Reclaimed", reclaim_amount=reclaim_amount.__format__(",d"))
+
+
+def _get_all_token_network_events(
```
hackaugusto, line 278: I changed the previous implementation which has one fetch request per user to a single request per token network. This has a few important benefits:

- We know that when multiple nodes interact with the same Geth/Parity node using the `eth_getLogs`interface the performance degrades very fast and leads to timeout errors. If we did use the previous implementation to fetch the logs from the nodes ,we could inadvertently cause other tests to crash because the underlying Ethereum node would be under stress, and the *others* requests could fail too.
- Additionally this return the event for close channels too, which is important, since calling withdraw on a closed channel is an error.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+    # target_block is never updated!
+    confirmation_blocks = DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS
+    latest_block = web3.eth.getBlock(BLOCK_ID_LATEST)
+    latest_block_number = latest_block["number"]
+    current_confirmed_head = BlockNumber(latest_block_number - confirmation_blocks)
+
+    # The block range is inclusive, meaning the block represented by
+    # last_fetched_block is skipped, so the block used has to be one before, to
+    # make sure the deployment block is included.
+    start_fetching_at = BlockNumber(token_network_deployed_at - 1)
```
hackaugusto, line 443: This was an off-by-one error that was not being properly handled before.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+    # executed, so we don't have to synchronize with the whole blockchain, just
+    # up to the last block used by the scenario run. Since we don't know the
+    # exact block, use the current confirmed head.
+    #
+    # This also assumes no RPC calls to pruned data will be performed, so the
+    # target_block is never updated!
+    confirmation_blocks = DEFAULT_NUMBER_OF_BLOCK_CONFIRMATIONS
+    latest_block = web3.eth.getBlock(BLOCK_ID_LATEST)
+    latest_block_number = latest_block["number"]
+    current_confirmed_head = BlockNumber(latest_block_number - confirmation_blocks)
```
hackaugusto, line 438: Caching the confirmed block, this saves RPC calls and streamlines the code (no need to do the same thing in multiple places).

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+            and event["participant1"] in address_to_candidate
+            and event["participant2"] in address_to_candidate
+        )
+        if is_useful_channel_open:
+            tracked_channels[event["channel_identifier"]] = event
+
+        is_useful_channel_close = event["event"] == ChannelEvent.CLOSED and (
+            event["channel_identifier"] in tracked_channels
+        )
+        if is_useful_channel_close:
```
hackaugusto, line 470: Handling closed channels to avoid invalid withdraws.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+    # There must be only one proxy per smart contract and and one JSONRPCClient
+    # per private key, this is necessary for transaction and nonce
+    # synchronization. This loops creates only the necessary instances. For
+    # this to work properly, this function cannot be executed concurrrently.
+    address_to_proxymanager: Dict[Address, ProxyManager] = dict()
+    for channel_open_event in tracked_channels.values():
+        for address in (channel_open_event["participant1"], channel_open_event["participant2"]):
+            if address not in address_to_proxymanager:
+                candidate = address_to_candidate[address]
+                client = JSONRPCClient(web3, candidate.privkey)
```
hackaugusto, line 484: This is the type of code that should be used by the higher layer, to enforce a single instance per address/privatekey.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,520 @@
...
+                    _withdraw_participant_left_capacity_from_channel,
+                    address_to_candidate=address_to_candidate,
+                    channel=channel_open_event,
+                    token_network=token_network,
+                )
+            )
+
+    # Wait until all transactions are mined, at this point we are ignoring
+    # errors.
+    pool.join()
```
hackaugusto, line 516: Changed to send all the transactions in parallel, this should have a big impact in wait time.

scenario_player/utils/reclaim.py
karlb: The ChannelOpened events are filtered by `participant1 == candidate`. Since you can't open channels with `participant1 == participant2` ti should be guaranteed, right?

hackaugusto: Humm, that is correct, but I had no way of telling this was the intention from the code, maybe a comment to point this out next time?

With that said, the code still had the problem of not handling withdraws for `participant2`, which I suppose was not intentional? And then this function would not be allowed to have that assumption.

karlb: All tokens are withdrawn by participant1, so there is no need to do a withdraw for participant2.  I tried to explain that in the docstring
>      All tokens will be withdrawn to participant1, ignoring all balance proofs. 
but apparently was not explicit enough.

scenario_player/utils/reclaim.py
```diff
@@ -0,0 +1,491 @@
...
+class ReclamationCandidate:
+    address: ChecksumAddress
+    keyfile_content: dict
+    node_dir: pathlib.Path
+
+    @cached_property
+    def privkey(self):
+        return decode_keyfile_json(self.keyfile_content, b"")
+
+    def get_client(self, web3: Web3) -> JSONRPCClient:
```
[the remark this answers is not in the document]
karlb: I didn't want to pass in `deploy` and `web3` into the "read account files from disk" part. But I agree that this is not ideal.

It should work fine for the current case though, so I'll let it be for now.

PR #552, opened 2020-05-05
To re-set the allowance two transactions are needed.

hackaugusto:


ulope:
`approve` / `transferFrom` is a gift that just keep on giving 🤦

PR #558, opened 2020-05-12
Raiden nodes use different eth nodes from list in enviroment file

### LINKED ISSUE #555: Every node in the SP should use a different geth node on mainet
Every node spawned by the Scenario Player should use a different Geth Node to avoid the syncing error

- [ ] Specify this issue
- [ ] Design solution

karlb:
This reduces the load on the eth nodes, making mainnet testing more
reliable. It can also help uncover synchronization problems that only
happen when raiden nodes connect to different eth nodes.

Closes #555


environment/mainnet.json
```diff
@@ -2,7 +2,23 @@
 	"environment_type": "production",
 	"matrix_server_list": "https://raw.githubusercontent.com/raiden-network/raiden-service-bundle/master/known_servers.main.test.yaml",
 	"pfs_with_fee": "https://pfs.rsb01.test.raiden.network",
-	"eth_rpc_endpoint": "http://geth.mainnet.ethnodes.brainbot.com:8545",
+        "eth_rpc_endpoints": [
```
palango, line 5: indention looks strange here

karlb: The usual tab vs spaces problem. Do we always want to use spaces in JSON? Then I'll change my config so that I'll get that right in future commits.

palango: No strong feelings from my side

scenario_player/utils/configuration/settings.py
```diff
@@ -15,6 +25,22 @@
 log = structlog.get_logger(__name__)
 
 
+@dataclass
+class EnvironmentConfig:
```
palango, line 29: Nice!

environment/mainnet.json
ulope: Personally I think tabs should be eradicated from time ;)

PR #563, opened 2020-05-12
Added time for synchronization

hackaugusto:
Adding wait time for every task with side effects. The time has to be large enough to account for a slow run, specially for parallel runs of all the scenarios that can be particularly stressful for the ethereum nodes and the services.

### DIFF
--- a/scenario_player/tasks/base.py
+++ b/scenario_player/tasks/base.py
@@ -39,6 +39,50 @@ _TASK_ID = 0
 class Task:
     _name: str
 
+    # The tasks in a scenario are written with the assumption of global
+    # consistency, which is wrong. Consider the following scenario:
+    #
+    #    - open_channel: {from: 0, to: 1, total_deposit: 1_000}
+    #    - open_channel: {from: 1, to: 2, total_deposit: 1_000}
+    #    - transfer: {from: 0, to: 2, amount: 1}
+    #
+    # There are multiple system runnning concurrently that must be synchronized for
+    # the `transfer` above to work:
+    #
+    # 1. The initiator (node 0) has to see the open event for the channel `1-2`.
+    # This is necessary since the initiator will verify that at least one route to
+    # the target exists, this is done to avoid sending unecessary IOUs to the PFS,
+    # without the route the payment fails (In general the initiator has to see all
+    # channel open events for every channel used in the path).
+    # 2. The node receiving a lock must see the deposit of the sender. In the case
+    # above node 1 has to see the deposit from node 0, and node 2 from node 1. If
+    # this is not satisfied a locked transfer message will be rejected. Note:
+    # Retries at the transport layer *do* fix this particular problem (In general,
+    # every payee (mediator or target) have to see the deposit of the payer).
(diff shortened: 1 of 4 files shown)


ulope:
I would suggest to move the `30`s default value to `scenario_player.tasks.api_base.RESTAPIActionTask` and set the value in `Task` to `0`.
All (Raiden) state changing is done through the node APIs, for which `RESTAPIActionTask` is the base.

Class hierarchy for illustration:
task_hierarchy

hackaugusto:
@ulope why?

ulope:
The more I think about this the less I like it (independent of my comment above).

The two main problems I see:
- Scenario run time will increase (in some cases significantly)
  - BF1, BF3 and BF4 will each take about an hour longer
- Scenario semantics change
  - Scenarios that perform many transfers in a row will suddenly stress the system a lot less (e.g. BF4)

I think we should revert this and instead fix the scenarios with proper explicit waits.

konradkonrad:
> Scenario semantics change
>
>   - Scenarios that perform many transfers in a row will suddenly stress the system a lot less (e.g. BF4)

Hummm, I did not think about that before. That is actually quite a price to pay :/

hackaugusto:
> Scenario run time will increase (in some cases significantly) 

Yes, that is the bug it fixes. The time it takes for state updates to propagate is unknown, and worse, it varies depending with the stress on the system. If we are using timeouts it has to be set to the worst case, otherwise we will have flakiness

>     Scenarios that perform many transfers in a row will suddenly stress the system a lot less (e.g. BF4)

this does not affect parallel runs, serial tasks would not be stressing the system anyways.

> I think we should revert this and instead fix the scenarios with proper explicit waits.

I dont mind it, I have always being vocal about using timeouts and how flaky they are, but it will take quite a bit of effort to build all the necessary primitives to wait for the correct state. Maybe first try to make a list of what are actual all the cases that need synchronization, and then estimate the time it takes to code all the necessary tasks?

ulope:
> Yes, that is the bug it fixes. The time it takes for state updates to propagate is unknown, and worse, it varies depending with the stress on the system. If we are using timeouts it has to be set to the worst case, otherwise we will have flakiness

My argument it that this is too blunt a solution to the problem.

> this does not affect parallel runs, serial tasks would not be stressing the system anyways.

I mean this specific case:

https://github.com/raiden-network/raiden/blob/6999c650235c975639b5d4d1582a628da460ee34/raiden/tests/scenarios/bf4_multi_payments_same_node.yaml#L63-L80

Before we would have constant concurrent transfers. With the forced wait this will devolve (due to unequal transfer times) into non-concurrent after only a couple of iterations.

Maybe let's talk about it during planning today.

Dominik1999:
The bf1 scenario took 4 hours tonight. Might that be related? 

`:parrot: [sp1] Scenario bf1_basic_functionality Successfull 4:10:52`

If so, I think we should not merge such a huge change in the last steps our testing for Alderaan. Let us get the scenarios on mainnet green now, re-run if a fail is caused due to a flaky state and merge this again after we have released?

PR #565, opened 2020-05-13
Fix casing in AssertPFSIoUTask

palango:


scenario_player/tasks/services.py
```diff
@@ -291,9 +291,9 @@ def _process_response(self, response_dict: dict):
         return response_dict
 
 
-class AssertPFSIoUTask(RESTAPIActionTask):
+class AssertPFSIouTask(RESTAPIActionTask):
     """
-    Assert PFS IoU task
+    Assert PFS Iou task
```
karlb, line 296: IOUs are usually written in all caps https://en.wikipedia.org/wiki/IOU. `AssertPFSIOUTask` is not very readable, so we could go with your approach, but I prefer the all caps version.

Issue #567, opened 2020-05-15
Use /status to check if client's API is up


palango:
In the clients logs I see lots of logs for the SP waiting for the API coming up. Instead we could use the new `/status` API in the client.
```
{"errors": "Service unavailable. Try again later.", "status_code": 503, "event": "Error processing request", "logger": "raiden.api.rest_utils", "level": "error", "timestamp": "2020-05-15 08:22:56.224145"}
```

PR #569, opened 2020-05-15
Use new raiden status endpoint

palango:
Resolves https://github.com/raiden-network/scenario-player/issues/567

~~I want to test this before getting this merged.~~
Worked fine in a testnet scenario.

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -441,7 +441,10 @@ class ScenarioRunner:
         log.info("Making sure all nodes have the same token network")
         self.ensure_token_network_discovery(token_proxy, token_network_address)
 
-        log.info("Setup done, running scenario", token_network_address=token_network_address)
+        log.info(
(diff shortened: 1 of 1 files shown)


scenario_player/runner.py
```diff
@@ -89,17 +89,17 @@ def is_udc_enabled(udc_settings: UDCSettingsConfig):
...
             while True:
                 try:
-                    if session.get(url).ok:
+                    if session.get(url).json()["status"] == "ready":
                         break
-                except requests.exceptions.RequestException:
+                except (requests.exceptions.RequestException, ValueError, KeyError):
                     pass
-                gevent.sleep(0.5)
+                gevent.sleep(1.0)
```
karlb, line 102: This makes me want to have a long-polling endpoint for the status, which delays returning the status when it is not ready. But for now, that would be over-engineering.

PR #571, opened 2020-05-19
Check enough token for all UDC deposits are available

palango:
Resolves https://github.com/raiden-network/raiden/issues/6169

For mainnet testing we're not minting tokens but use the ones in the existing accounts.
In the case of not enough token available for all nodes to be funded, we ran into the issue above.

This check should fix that issue. It is however not applicable to the "normal" scenario player, as the minting would be done in the next step and it will fail in every case.


karlb:
Looks good, but we can't merge this into master until we have a proper mint switch.

PR #583, opened 2020-06-30
Fix handling of non-dict configs

palango:
As seen in `AttributeError: 'int' object has no attribute 'get'`

### DIFF
--- a/scenario_player/tasks/base.py
+++ b/scenario_player/tasks/base.py
@@ -110,9 +110,12 @@ class Task:
         self._runner.running_task_count += 1
         self._start_time = time.monotonic()
         try:
+            timeout_s = None
+            # Config can be something else than a dictionary
+            if isinstance(self._config, dict):
+                timeout_s = self._config.get("timeout", self.DEFAULT_TIMEOUT)
             # Zero means no timeout is desired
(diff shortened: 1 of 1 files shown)


scenario_player/tasks/base.py
```diff
@@ -110,9 +110,12 @@ def __call__(self, *args, **kwargs):
         self._runner.running_task_count += 1
         self._start_time = time.monotonic()
         try:
+            timeout_s = None
+            # Config can be something else than a dictionary
```
hackaugusto, line 114: when would the config be something else? this seems a problem that should be handled in the caller site, and not here.

palango: For example in the `WaitTask`, the config is just an `int`

hackaugusto: The point is that it should be fixed there then.

palango: I guess we want to support this:
```
 - wait: 30
```
instead of 
```
- wait: {time: 30}
```

palango: I propopose to fix that later: https://github.com/raiden-network/scenario-player/issues/584

PR #588, opened 2020-07-24
Provide an example with PFS usage

karlb:
To use this, run a local PFS and set the correct claims file, e.g.
`--claims-file ~/.raiden/scenario-player/scenarios/raiddit-pfs/claims.jsonl`

Also reduce default-reveal-timeout in raiddit.yaml to 10 to fix
reveal/settle timeout problems.


karlb:
Closed by https://github.com/raiden-network/raiden/pull/6420

PR #597, opened 2020-07-29
Direct channels

palango:


scenario_player/runner.py
```diff
@@ -658,9 +658,10 @@ def setup_claims(self, token_network_address: TokenNetworkAddress) -> None:
                 Address(os.urandom(20))
                 for _ in range(claims_config.additional_address_count - len(self.node_controller))
             ]
-            node_addresses = {
-                to_canonical_address(address) for address in self.node_controller.addresses
-            }
+            node_addresses = [
```
ulope, line 661: Ah you need them in order...

PR #598, opened 2020-07-30
Fix free port generation under high concurrency

ulope:
When a lot of nodes are started simultaneously just asking the OS for a
free port can lead to conflicts due to port reuse.

This fixes that. See comment in the code for details.

### DIFF
--- a/scenario_player/node_support.py
+++ b/scenario_player/node_support.py
@@ -201,7 +201,9 @@ class NodeRunner:
         _ = self._keystore_file  # noqa: F841
         _ = self._raiden_bin  # noqa: F841
 
-    def start(self):
+    def start(self, wait: bool = False):
+        from scenario_player.runner import wait_for_nodes_to_be_ready
+
         log.info(
             "Starting node",
             node=self._index,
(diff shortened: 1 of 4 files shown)


scenario_player/node_support.py
```diff
@@ -346,11 +348,26 @@ def api_address(self):
...
+                    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+                    sock.bind(("127.0.0.1", 0))
+
+                    sock_addr = sock.getsockname()
+                    port = sock_addr[1]
+
+                    # Connect to the socket to force it into TIME_WAIT state (see
+                    # above)
+                    sock.listen(1)
+                    sock2 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
hackaugusto, line 366: why not use the same function available in the raiden repo? this approach was **not** reliable on CI, which was a cause of flakiness, I'd assume that it will be flaky in the scenario servers too

ulope: This is only for the raiddit branch.

PR #601, opened 2020-09-23
Fix smoketest

ulope:
Blocked-by: raiden-network/raiden#6561

- Upgrade Synapse dev requirement in order to catch up with Raiden.
- The smoketest was using one of the public test Matrix servers instead of the local one.


fredo (commented):
LGTM

PR #602, opened 2020-09-28
Use private routing in smoketest

karlb:
The local routing is about to be removed. We can just use `private`
routing in the smoketest, instead.


palango (approved):
Change looks good, CI tricky 🤯

Issue #593, opened 2020-07-28
raiden_version: support --raiden-version cli param and/or raiden_version environment to overwrite scenario's raiden_version


andrevmatos:
Analysis
=======
Right now, if one wants to test Raiden with a different version/binary/executable, one needs to change the scenario file and set the `raiden_version` directly there. It'd be great to have it overwritten by a `--raiden-version` command line option and/or a `raiden_version` environment entry in the environment file, in order to easily test with a custom raiden version without the need to change the scenario itself.
This is needed in order to run the Raiden scenarios with the Light Client.

christianbrb:
@ulope @Dominik1999 Any remarks or could we go ahead and implement this?

PR #603, opened 2020-09-30
Add options to specify raiden client executable to use

palango:
Fixes https://github.com/raiden-network/scenario-player/issues/593

This PR removes the hardcoded `raiden_version` in the scenario file and replaces it with options to set the executable to use from the CLI, the environment file or overwrite on a per-node basis in the scenario.

This also makes it possible to remove the `RaidenReleaseKeeper` class. As `flake8` complained about some code this also fixes some warnings not strictly related to the functionality.


scenario_player/main.py
```diff
@@ -323,6 +328,9 @@ def run_(
             )
         notify_tasks_callable = post_task_state_to_rc
 
+    # TODO: make this used
+    print(notify_tasks_callable)
```
fredo, line 332: is this intended to stay?

scenario_player/node_support.py
```diff
@@ -68,109 +56,16 @@
...
-    def __init__(self, runner: "ScenarioRunner", index: int, raiden_version, options: dict):
+    def __init__(self, runner: "ScenarioRunner", index: int, raiden_client: str, options: dict):
         self._runner = runner
         self._index = index
-        self._raiden_version = raiden_version
         self._options = options
         self._nursery: Optional[Nursery] = None
+
+        # Overwrite raiden client to use, if given in options
+        self._raiden_client = options.pop("raiden-client", raiden_client)
```
fredo, line 67: Just to assure, is option["raiden-client"] used somewhere afterwards? As it pops it from the dict.

fredo (commented):
In general looks good to me, left some remarks. 

One question pops up. When using the executable, if we would like to show the current version we could query it from the CLI, right? I saw that some logs showing the version used were removed. Would be nice to see it somewhere.

scenario_player/node_support.py
palango: No, it isn't

scenario_player/main.py
palango: It was passed to `orchestrate` before, but never used afterwards. Maybe @ulope can shine some light on this?

examples/scenario-example-v2.yaml
```diff
@@ -72,6 +69,7 @@ nodes:
   #node_options:
   #  0:
   #    matrix-server: https://transport01.raiden.network
+  #    raiden-client: light-client
```
ulope, line 72: Hm mixing in this "meta" option into the cli options feels a bit "unclean".

scenario_player/node_support.py
```diff
@@ -516,7 +417,6 @@ def __init__(
         if config.restore_snapshot:
             if self.snapshot_manager.restore():
                 self.snapshot_restored = True
-        log.info("Using Raiden version", version=config.raiden_version)
```
ulope, line 519: Maybe keep this and log the full path to the executable that is being used?

ulope (approved):
LGTM.

I'm still a bit sad that we lose the ability to run different versions without external tooling.
But nobody ever using that feature speaks pretty clearly that it wasn't very useful.

scenario_player/main.py
ulope: The idea of that callback was to have a way of notifying external tools (e.g. rocketchat) about the task state. It's very possible that this got broken in one of the refactorings along the way.

examples/scenario-example-v2.yaml
palango: I know what you mean. Any other ideas? 

I'd also be ok with not having this feature at this point and implementing a "proper" solution later.

scenario_player/main.py
palango: Ok, then I create an issue to bring it back when we need it

PR #605, opened 2020-09-30
Update dev requirements

palango:


pyproject.toml
```diff
@@ -55,19 +55,18 @@ keywords = "raiden, raiden-network"
...
+    "flake8",
+    "isort",
     "matrix-synapse==1.19.1",
-    "mypy==0.782",
+    "mypy",
     "pylint",
     "pytest-cov",
-    "pytest-dependency==0.4.0",
-    "responses==0.10.6",
+    "responses",
```
karlb, line 69: Is it a good idea not to pin the versions?

palango: Brought back the pins.

karlb (approved):

Issue #607, opened 2020-10-01
Error on render of BF6 scenario


andrevmatos:
Description
=========
Current master. Getting the traceback below trying to run BF6 (exactly as on raiden repo), using provided `development.json` environment and replacing `raiden_client` with LC's CLI script:
```
Traceback (most recent call last):
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/scenario_player/main.py", line 347, in run_
    scenario_runner = ScenarioRunner(
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/scenario_player/runner.py", line 267, in __init__
    self.definition = ScenarioDefinition(scenario_file, data_path, self.environment)
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/scenario_player/definition.py", line 31, in __init__
    rendered_yaml = yaml_template.render(**asdict(environment))
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/mnt/sec/Dev/brainbot/sp/venv/lib/python3.8/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 1, in <module>
jinja2.exceptions.UndefinedError: list object has no element 1
```

Expected Behavior
===============
No error, scenario runs

karlb:
This happens when the scenario assigns specific matrix servers to the nodes, but the scenario does not contain any information about matrix servers. The SP could
* look at the `environment` key and fetch the default server list like Raiden itself does
* or just use auto for all nodes
* or we require matrix server information in the environment file

I prefer the first option, since the scenario explicitly asks for different servers.

palango:
I only see the `matrix_server_list` key missing in `development.json`. Isn't adding it the easiest solution short term?

ulope:
Yes having the SP load the list sounds like the cleanest option. 
[...]

PR #609, opened 2020-10-02
Always load a server list

karlb:
Some scenarios explicitly assign different matrix servers to raiden
nodes. For this to work, the SP needs a list of matrix serves. When
none is given, use the default server list for the given `environment`
key in the environment JSON file.

This undoes part of https://github.com/raiden-network/scenario-player/commit/7c051a12dd25d97d64b0687dc16620bd8f53de16

Closes https://github.com/raiden-network/scenario-player/issues/607.

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -11,7 +11,7 @@ from io import StringIO
 from itertools import cycle, islice
 from pathlib import Path
 from tempfile import mkdtemp
-from typing import IO, List, Sequence, Union
+from typing import IO, List, Sequence
 
 import click
 import gevent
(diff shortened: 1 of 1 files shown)


scenario_player/main.py
```diff
@@ -269,12 +269,13 @@ def _load_environment(environment_file: IO) -> EnvironmentConfig:
...
-    if matrix_server_list is not None:
-        matrix_servers = [URI(uri) for uri in get_matrix_servers(matrix_server_list)]
-        if len(matrix_servers) < 4:
-            matrix_servers = list(islice(cycle(matrix_servers), 4))
+    matrix_server_list = environment.pop(
+        "matrix_server_list",
+        DEFAULT_MATRIX_KNOWN_SERVERS[Environment(environment["environment_type"])],
+    )
+    matrix_servers: Sequence[URI] = get_matrix_servers(matrix_server_list)  # type: ignore
+    if len(matrix_servers) < 4:
```
palango, line 277: Where is the 4 coming from?

karlb: This is the highest number of servers any existing scenario requires. Neither elegant nor flexible, but works for now. This is also not new in this PR.

palango (approved):
Thanks

ulope (commented):
Due to 

https://github.com/raiden-network/scenario-player/blob/ae6bafe7ef888f29aa47abafeaff48370303ac8c/scenario_player/definition.py#L41-L45

this will now again always force a fixed matrix server on all nodes in all scenarios. 
I don't think this is a good idea.

But as a short term fix it's fine I guess.

karlb:
> Due to
> 
> https://github.com/raiden-network/scenario-player/blob/ae6bafe7ef888f29aa47abafeaff48370303ac8c/scenario_player/definition.py#L41-L45
> 
> this will now again always force a fixed matrix server on all nodes in all scenarios.
> I don't think this is a good idea.
> 
> But as a short term fix it's fine I guess.

We could still explicitly set "auto" in the scenario, can't we?

We probably need to draw a decision matrix at some point to know what we actually want to do in each case and the redo this server selection cleanly according to that.

PR #613, opened 2020-10-09
Fix delayed returns from tasks

### LINKED ISSUE #611: AssertPFSIOU task keeps retrying until 5min timeout even if it's succeeding
Description
=========
It eventually succeeds, but takes 5min timeout, even though the request condition is fulfilled in the first request.
I didn't notice this behavior on other tasks, only on `AssertPFSIOU`

Task:
```
AssertPFSIOU 0:04:04.681630: {'source': 3, 'amount': 500000000000000000}
```

Logs filled with:
```
{"url": "https://pfs.transport01.raiden.network/api/v1/_debug/ious/0xA0661172D1aD4eDf76fDF3fAd2bE5F49ac9f4425", "method": "get", "json": {}, "event": "Requesting", "logger": "scenario_player.tasks.api_base", "level": "debug", "greenlet_name": "orchestration", "timestamp": "2020-10-08 13:45:40.442408"}
{"json": {"sender": "0xA0661172D1aD4eDf76fDF3fAd2bE5F49ac9f4425", "amount": 500000000000000000, "expiration_block": 3739450}, "event": "Received response", "logger": "scenario_player.tasks.api_base", "level": "debug", "greenlet_name": "orchestration", "timestamp": "2020-10-08 13:45:40.607931"}
{"ex": "None", "event": "Assertion failed, retrying...", "logger": "scenario_player.tasks.base", "level": "debug", "greenlet_name": "orchestration", "timestamp": "2020-10-08 13:45:40.609282"}
```

Return of that `_debug` URL request:
```json
{
    "amount": 500000000000000000,
    "expiration_block": 3739450,
    "sender": "0xA0661172D1aD4eDf76fDF3fAd2bE5F49ac9f4425"
}
```

Expected Behavior
===============
The task succeeds the first time the URL returns the expected value.

palango:
Those tasks only threw in case of errors, but didn't return a value in
the happy case, which led to additional iterations.

Fixes #611

### DIFF
--- a/scenario_player/tasks/services.py
+++ b/scenario_player/tasks/services.py
@@ -341,8 +341,7 @@ class AssertPFSIOUTask(RESTAPIActionTask):
         else:
             source_address = self._runner.get_node_address(source)
 
-        params = dict(pfs_url=pfs_url, source_address=source_address)
-        return params
+        return dict(pfs_url=pfs_url, source_address=source_address)
 
     def _process_response(self, response_dict: dict):
 
@@ -358,3 +357,4 @@ class AssertPFSIOUTask(RESTAPIActionTask):
                 raise ScenarioAssertionError(
                     f"Expected amount of {exp_iou_amount} but got {actual_iou_amount}."
                 )
+        return response_dict
--- a/scenario_player/tasks/base.py
+++ b/scenario_player/tasks/base.py
@@ -126,11 +126,11 @@ class Task:
                                 return_val = self._run(*args, **kwargs)
                             except ScenarioAssertionError as ex:
                                 exception = ex
+                                log.debug("Assertion failed, retrying...", ex=str(exception))
 
                             if return_val:
                                 break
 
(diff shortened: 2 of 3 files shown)


scenario_player/tasks/services.py
```diff
@@ -358,3 +357,4 @@ def _process_response(self, response_dict: dict):
                 raise ScenarioAssertionError(
                     f"Expected amount of {exp_iou_amount} but got {actual_iou_amount}."
                 )
+        return response_dict
```
palango, line 360: We should add type annotations for this. 

@ulope Do/should those always return dicts, or can it be any value?

ulope: IIRC `_process_response` should always return a dict because another subclass might want to further process it.

ulope (approved):
LGTM

Nice catch! 
How did we never notice that before?

scenario_player/tasks/services.py
palango: I'll fix that in a follow-up.

PR #619, opened 2020-10-13
Fix more retry handling

### LINKED ISSUE #617: Retry broken for AssertMSClaimTask
Description
=========

The retry logic is broken for the `AssertMSClaimTask`. It retries until the timeout is reached, even though the condition is fulfilled before.

Expected Behavior
===============

Should finish task after condition is fulfilled.

palango:
Fixes #617


karlb (approved):
So the assert retries never worked as intended?

palango:
> So the assert retries never worked as intended?

The retries worked, but they never stopped before the timeout. SO even if successful, the task always took 5mins.

Issue #610, opened 2020-10-02
Clean up package management


palango:
The current package management is a mess. We need to clean it up.

- [ ] Remove `flit`
- [ ] Use different dependency management. In the current `pyproject.toml` we cannot declare that we need a git raiden installation
  - use `pip-tools` as well?

PR #621, opened 2020-10-13
Use poetry for package management

palango:
Fixes https://github.com/raiden-network/scenario-player/issues/610

The current state of package management was a mess and people got confused on how to install the SP correctly. This fixes that by using poetry to manage the package. This makes it possible to mark raiden as a git dependency and the CI is simplified as well.

@ulope @karlb As this brings in another tool (besides `setup.py` and `pip-tools`) please let me know if you think we should go this way. If so, I'll clean this up a bit more.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,14 +42,8 @@ commands:
     description: Restore cache and virtualenv, and install dependencies.
     steps:
       - restore_cache:
-          key: pip-cache
-
-      - restore_cache:
           keys:
             - deps-{{ checksum "poetry.lock" }}
-
-      - restore_cache:
-          key: smoketest-cache
 
       - install_tools_for_ci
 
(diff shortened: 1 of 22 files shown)


ulope:
I think this repo is a good candidate to try out new tools like this.

Two things that come to mind:
- Does this still support installing Raiden in editable mode?
- The `Makefile` `install` / `install-dev` targets should probably be updated to use `poetry` directly instead of going via pip

palango:
> Does this still support installing Raiden in editable mode?

The only thing I found was `path` dependencies that have a `develop` option which installs the dependency in editable mode: https://python-poetry.org/docs/dependency-specification/#path-dependencies

ulope:
> > Does this still support installing Raiden in editable mode?
> 
> The only thing I found was `path` dependencies that have a `develop` option which installs the dependency in editable mode: https://python-poetry.org/docs/dependency-specification/#path-dependencies

Hm yeah but we probably don't want to assume a filesystem layout.

This does seem to work though:
```shell
$ poetry install
[...]
$ poetry run pip install -Ue ../raiden
[...]
Installing collected packages: raiden
  Found existing installation: raiden 1.1.1.dev270+g4b910649d
    Uninstalling raiden-1.1.1.dev270+g4b910649d:
      Successfully uninstalled raiden-1.1.1.dev270+g4b910649d
  Running setup.py develop for raiden
Successfully installed raiden
```

karlb:
I generally like poetry, but it might not be really mature yet. Since you already did the work and the git dependencies sound great, let's just give it a try!

ulope (approved):

PR #622, opened 2020-10-19
Re-add removed `install-dev` make target

karlb:
This is used by the Raiden SP smoke test. I thought about changing the
call in Raiden, but I like to have the consistency of having an
`install-dev` target in all repos, even if it is the same as `install`
for some.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,8 @@ mypy:
 install:
 	poetry install
 
+install-dev: install
+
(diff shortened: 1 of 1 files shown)


palango (approved):

karlb:
This is not really necessary with https://github.com/raiden-network/raiden/pull/6616#pullrequestreview-511697203, but I still like it.

PR #635, opened 2020-10-30
Update event filtering after raiden update

karlb:
The selective event filtering introduced in
https://github.com/raiden-network/raiden/pull/6563 changed the API,
which make this commit necessary.


palango (approved):
Great! Thanks!

PR #637, opened 2020-11-10
Increase connection pool size

### LINKED ISSUE #554: Evaluate the size of the connection pool necessary by the scenario player
Running the scenario bf7 will spam the screen with messages:

> 2020-05-07 09:26:14.716482 [warning  ] Connection pool is full, discarding connection: 4.geth.mainnet.ethnodes.brainbot.com [urllib3.connectionpool]

This is not only distracting, but it can also be misleading and perform poorly. The size of connection pool should be set prior to executing the scenarios to reuse the connections and avoid the warnings.

palango:
Fixes #554

### DIFF
--- a/scenario_player/runner.py
+++ b/scenario_player/runner.py
@@ -52,6 +52,7 @@ from scenario_player.definition import ScenarioDefinition
 from scenario_player.exceptions import ScenarioError, TokenNetworkDiscoveryTimeout
 from scenario_player.node_support import NodeController, NodeRunner
 from scenario_player.utils import TimeOutHTTPAdapter
+from scenario_player.utils.configuration.nodes import NodesConfig
 from scenario_player.utils.configuration.settings import (
     EnvironmentConfig,
     SettingsConfig,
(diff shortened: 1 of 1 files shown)


scenario_player/runner.py
```diff
@@ -152,12 +153,19 @@ def determine_run_number(scenario_dir: Path) -> int:
...
+def make_session(auth: str, settings: SettingsConfig, node_config: NodesConfig) -> Session:
+    num_connections = node_config.count * 10
+
     session = Session()
     if auth:
         session.auth = cast(Tuple[str, str], tuple(auth.split(":")))
-    session.mount("http", TimeOutHTTPAdapter(timeout=settings.timeout))
-    session.mount("https", TimeOutHTTPAdapter(timeout=settings.timeout))
+    session.mount(
+        "http", TimeOutHTTPAdapter(timeout=settings.timeout, pool_connections=num_connections)
```
palango, line 163: @ulope Any reason why we instantiate two `TimeOutHTTPAdapter`s here? Would one be enough?

ulope: Not sure tbh. The example in the requests documentation was doing it this way.

ulope (approved):
LGTM

PR #638, opened 2020-11-11
Adapt makefile and CI config to use poetry

ulope:
Automatically invoke targets in the makefile through poetry

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ commands:
             tar zxf geth.tar.gz
             cd geth*/
             install -m 755 geth ~/.local/bin/geth
+            pip install -U poetry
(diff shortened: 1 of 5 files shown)


palango (approved):
Thanks for fixing this. lgtm.

Regarding the CI failure. Should we remove the lock file form the repo to make it pass those minor updates?

palango:
> Thanks for fixing this. lgtm.
> 
> Regarding the CI failure. Should we remove the lock file form the repo to make it pass those minor updates?

@ulope Opinions on this?

ulope:
Err, strange I thought I already commented here.

I wouldn’t remove the lockfile. Otherwise we’ll be back to non-reproducible builds and could just have kept using pip.

ulope:
Thanks for fixing the PR @palango!

PR #648, opened 2021-01-12
Use geth for nightly runs as well

palango:
Parity shows some problems, let's switch to `geth` for now.


andrevmatos (approved):
lgtm, ty

palango:
Force-merging to see effect in nightlies.

PR #655, opened 2021-01-25
setup: Increase default UDC deposit

### LINKED ISSUE #276: Increase default UDC deposit
Analysis
---------------
The current default UDC deposit is to low

Problem
-----------
It might fail

Solution
----------
Increase UDC deposit limit

palango:
This is set to a value high enough to pay for a monitoring request, because
the light client disables receiving of payments otherwise.
See https://github.com/raiden-network/light-client/issues/2524

Fixes #276

### DIFF
--- a/scenario_player/utils/configuration/settings.py
+++ b/scenario_player/utils/configuration/settings.py
@@ -123,8 +123,13 @@ class UDCTokenSettings:
 
     @property
     def balance_per_node(self) -> int:
-        """The required amount of UDC/RDN tokens required by each node."""
-        return int(self.dict.get("balance_per_node", 50 * self.environment.pfs_fee))
+        """The required amount of UDC/RDN tokens required by each node.
+
+        This is set to a value high enough to pay for a monitoring request, because
+        the light client disables receiving of payments otherwise.
+        See https://github.com/raiden-network/light-client/issues/2524
+        """
+        return int(self.dict.get("balance_per_node", 200 * self.environment.pfs_fee))
 
     @property
     def max_funding(self) -> int:
--- a/tests/unittests/utils/configuration/test_settings.py
+++ b/tests/unittests/utils/configuration/test_settings.py
@@ -129,7 +129,7 @@ class TestUDCSettingsConfig:
 class TestUDCTokenConfig:
     @pytest.mark.parametrize(
         "key, expected",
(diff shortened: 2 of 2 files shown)


andrevmatos:
There're [scenarios which explicitly require a lower UDC deposit](https://github.com/raiden-network/raiden/blob/02284088976464377da28d89668f40188f22763e/raiden/tests/scenarios/ms4_udc_too_low.yaml#L13). Are we sure this isn't implicitly required by some of the other scenarios too?
In either way, couldn't this just mask/hide the issue about a slightly differing behavior between clients? Should we pursue a consensus on this rule, or are we fine to rule it as an implementation detail?
PS: The behavior difference is that the LC has an additional security feature where it disables receiving (both on capabilities and internally) when there's not enough UDC deposit to back the configured `monitoringReward`, while PC ignores this and always accepts incoming transfers.

palango:
> There're scenarios which explicitly require a lower UDC deposit. Are we sure this isn't implicitly required by some of the other scenarios too?

I'd say in this case the scenarios **must** set the setting, because otherwise the "safe" default will be used.

> In either way, couldn't this just mask/hide the issue about a slightly differing behavior between clients? Should we pursue a consensus on this rule, or are we fine to rule it as an implementation detail?
PS: The behavior difference is that the LC has an additional security feature where it disables receiving (both on capabilities and internally) when there's not enough UDC deposit to back the configured monitoringReward, while PC ignores this and always accepts incoming transfers.

That's a good point. Let's ping @karlb and I will push for some answer of him and the PC team.

karlb:
> > PS: The behavior difference is that the LC has an additional security feature where it disables receiving (both on capabilities and internally) when there's not enough UDC deposit to back the configured monitoringReward, while PC ignores this and always accepts incoming transfers.

We don't do this intentionally to keep the minimum UDC deposit required for Raiden usage low. Both approaches seem reasonable and having consistency between the clients is generally preferable. As long as business/legal agree, I'm fine with either approach.

andrevmatos (approved):
lgtm

karlb:
We talked about this in today's weekly meeting. The expectation for the python client is that it is usually online, thus reducing the importance for monitoring compared to the light client. Therefore, we keep the current behavior of allowing receiving of payments without monitoring in the python client. Differing behavior between the clients is deemed acceptable in this case.

PR #657, opened 2021-02-11
Change RaidenTestSetup.arg's `password_file` type

ezdac:
Raiden's commit 5911b5a0 changes the type of
`password_file` in RaidenTestSetup.args from a click.Path
to a path.
Now this is mirrored in the access of that value.

After this is merged, SP runs will probably fail until https://github.com/raiden-network/raiden/pull/6819 is merged. Since Raiden PR #6819 depends on this PR to be included in the SP, there is a circular dependency between the two PRs.

### DIFF
--- a/scenario_player/main.py
+++ b/scenario_player/main.py
@@ -559,7 +559,7 @@ def smoketest(eth_client: EthClient):
                 config_file = create_smoketest_config_file(setup, datadir)
 
                 keystore_file = os.path.join(setup.args["keystore_path"], "keyfile")
-                password_file = setup.args["password_file"].name
+                password_file = setup.args["password_file"]
                 print_step("Running scenario player")
                 append_report("Scenario Player Log", captured_stdout.getvalue())
                 env = EnvironmentConfig(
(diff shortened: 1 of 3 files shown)


scenario_player/main.py
```diff
@@ -559,7 +559,7 @@ def smoketest(eth_client: EthClient):
                 config_file = create_smoketest_config_file(setup, datadir)
 
                 keystore_file = os.path.join(setup.args["keystore_path"], "keyfile")
-                password_file = setup.args["password_file"].name
+                password_file = setup.args["password_file"]
```
ulope, line 562: Is this correct?
If it's now a `Path` then this should probably be `str(setup.args["password_file"])`, or the downstream needs to be changed to work with Path instances.

ezdac: The downstream (`get_password( )` ) uses `open(password_file, 'r' )` to read in the file, which should be fine to accept `Path` instances?

ulope: Hm yeah since the `__fspath__` protocol was introduced it should indeed work.

ulope (approved):

PR #663, opened 2021-03-26
Use unstable contract env for development env

karlb:
Part of https://github.com/raiden-network/light-client/issues/2521

Merge after https://github.com/raiden-network/raiden-services/pull/951 and configuring `*.services-dev.raiden.network/` to use the unstable env.

### DIFF
--- a/scenario_player/environment/development.json
+++ b/scenario_player/environment/development.json
@@ -1,5 +1,6 @@
 {
     "environment_type": "development",
+    "development_environment": "unstable",
     "pfs_with_fee": "https://pfs.transport01.raiden.network",
     "eth_rpc_endpoints": [
         "http://parity.goerli.ethnodes.brainbot.com:8545"
--- a/scenario_player/utils/configuration/settings.py
+++ b/scenario_player/utils/configuration/settings.py
@@ -5,6 +5,7 @@ from typing import Callable, Iterator, List, Optional, Sequence, Union
 
 import structlog
 from eth_typing import URI
+from raiden_contracts.contract_manager import ContractDevEnvironment
 from typing_extensions import Literal
 
 from raiden.utils.typing import (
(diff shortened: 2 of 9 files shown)


scenario_player/utils/configuration/settings.py
```diff
@@ -40,9 +41,11 @@ class EnvironmentConfig:
     raiden_client: str
     wait_short: int
     wait_long: int
+    development_environment: ContractDevEnvironment = ContractDevEnvironment.DEMO
 
     def __post_init__(self):
         self.eth_rpc_endpoint_iterator = itertools.cycle(self.eth_rpc_endpoints)
+        self.development_environment = ContractDevEnvironment(self.development_environment)
```
palango, line 48: Is this explicit cast necessary?

scenario_player/environment/development.json
```diff
@@ -1,5 +1,6 @@
 {
     "environment_type": "development",
+    "development_environment": "unstable",
```
palango, line 3: Should we define this in the other files as well, just for consistency?

palango (approved):

scenario_player/utils/configuration/settings.py
karlb: This allows calling `EnvironmentConfig(**json.load(...))` directly. Without it, you would have to explicitly convert the "unstable" string to `EnvironmentConfig("unstable")`, first. That would work too, but this way the JSON file and the EnvironmentConfig can be seen as the same thing more easily.

scenario_player/environment/development.json
karlb: Since the default choice is the right one for all environments but this one, I prefer to leave it out. It could just cause confusion with the similarly named "environment_type".

PR #674, opened 2021-05-14
Add Makefile target to install local raiden

karlb:
This make it easier to run arbitrary raiden version on your local
machine. The path is hardcoded and might not work for everyone, but it
will be a helpful documentation even in that case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,14 @@ install: have-poetry
 install-dev: have-poetry
 	poetry install
 
+install-local-raiden:
+	echo "This only works when Raiden is installed in ../raiden"
+	echo "If pip install -e ."
+	sed -i 's/^raiden = {.*/raiden = { path = "..\/raiden"}/' pyproject.toml
+	poetry update raiden
+	poetry install
+	echo "If you get pkg_resources.ContextualVersionConflict, you need to update your egg-info by calling 'pip install -e ../raiden'."
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,6 +33,14 @@ install: have-poetry
 install-dev: have-poetry
 	poetry install
 
+install-local-raiden:
+	echo "This only works when Raiden is installed in ../raiden"
+	echo "If pip install -e ."
+	sed -i 's/^raiden = {.*/raiden = { path = "..\/raiden"}/' pyproject.toml
+	poetry update raiden
+	poetry install
+	echo "If you get pkg_resources.ContextualVersionConflict, you need to update your egg-info by calling 'pip install -e ../raiden'."
```
ulope, line 42:
```suggestion
ifeq (, $(wildcard ../raiden/.git))
	echo "This only works when a Raiden source checkout is located in ../raiden"
else
	sed -i 's/^raiden = {.*/raiden = { path = "..\/raiden"}/' pyproject.toml
	poetry update raiden
	poetry install
	echo "If you get pkg_resources.ContextualVersionConflict, you need to update your egg-info by calling 'pip install -e ../raiden'."
endif
```

ulope (approved):
Added a suggestion with a check to only run the commands if a local checkout exists in the expected path.

PR #683, opened 2021-06-02
Use geth node on goerli again

palango:
This was disabled when there were problems with out geth node and never
reenabled.


andrevmatos (approved):
lgtm; if parity node is behaving weirdly, some scenarios may still fail

PR #695, opened 2021-08-19
Send SIGUSR1 to client process on Timeout

konradkonrad:
This will send `SIGUSR1` to all client processes when a task hits a
timeout. The intention is to trigger a debugging hook, installed in
the python client, to help with deadlock detection.

See
https://github.com/raiden-network/raiden/issues/7141#issuecomment-901757801
for reference.

### DIFF
--- a/tests/unittests/conftest.py
+++ b/tests/unittests/conftest.py
@@ -1,3 +1,4 @@
+import sys
 from collections import defaultdict
 from typing import Dict
 from unittest.mock import MagicMock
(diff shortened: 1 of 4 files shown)


tests/unittests/conftest.py
```diff
@@ -104,6 +105,9 @@ def __len__(self):
     def address_to_index(self) -> Dict[ChecksumAddress, int]:
         return {runner.address: i for i, runner in enumerate(iter(self))}  # type: ignore
 
+    def send_debugging_signal(self) -> None:
+        sys.stderr.write("Debugging signal sent\n")
```
netcriptus, line 109: Should the debug message really go to `stderr`? I'm not sure I understand why.

netcriptus (approved):
One minor comment, overall LGTM

ulope:
How does the light client react to `USR1`?

konradkonrad:
> How does the light client react to `USR1`?

Probably NOOP reaction at all? @andrevmatos ?

andrevmatos:
We don't have specific hooks for USR1. Nodejs does, by enabling inspector on the process, but without a DevTool to connect to it, it should be noop for us.

PR #697, opened 2021-08-24
Add AssertChannelSettledEvent

ezdac:
Since the call-arguments for `participant1` and `participant2` are
unordered and the order is not enforced based on who calls the function,
both possibilities have to be checked when making assertions on
ChannelSettled events that are coop-settle related.
This commit introduces the `AssertChannelSettledEvent` task,
that allows for more advanced assertions related to ChannelSettled
events.

### DIFF
--- a/scenario_player/tasks/blockchain.py
+++ b/scenario_player/tasks/blockchain.py
@@ -86,6 +86,14 @@ def query_blockchain_events(
     ]
 
 
+def verify_config(config, required_keys):
+    all_required_options_provided = all(key in config.keys() for key in required_keys)
+    if not all_required_options_provided:
+        raise ScenarioError(
+            "Not all required keys provided. Required: " + ", ".join(required_keys)
+        )
+
+
 class AssertBlockchainEventsTask(Task):
     """Assert on blockchain events.
 
@@ -122,16 +130,10 @@ class AssertBlockchainEventsTask(Task):
     ) -> None:
         super().__init__(runner, config, parent)
 
-        required_keys = ["contract_name", "event_name", "num_events"]
-        all_required_options_provided = all(key in config.keys() for key in required_keys)
-        if not all_required_options_provided:
-            raise ScenarioError(
-                "Not all required keys provided. Required: " + ", ".join(required_keys)
-            )
-
+        verify_config(config, required_keys=["contract_name", "event_name"])
         self.contract_name = config["contract_name"]
         self.event_name = config["event_name"]
-        self.num_events = config["num_events"]
+        self.num_events = config.get("num_events", None)
         self.event_args: Dict[str, Any] = config.get("event_args", {}).copy()
 
         self.web3 = self._runner.client.web3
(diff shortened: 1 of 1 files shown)


scenario_player/tasks/blockchain.py
```diff
@@ -122,16 +130,10 @@ def __init__(
...
-        if not all_required_options_provided:
-            raise ScenarioError(
-                "Not all required keys provided. Required: " + ", ".join(required_keys)
-            )
-
+        verify_config(config, required_keys=["contract_name", "event_name"])
         self.contract_name = config["contract_name"]
         self.event_name = config["event_name"]
-        self.num_events = config["num_events"]
+        self.num_events = config.get("num_events", None)
```
istankovic, line 136: Don't need `None` as the second argument to `get`, it's the default.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
             events = [e for e in events if e["args"] and event_args_items & e["args"].items()]
 
-        # Raise exception when events do not match
-        if not self.num_events == len(events):
-            raise ScenarioAssertionError(
-                f"Expected number of events ({self.num_events}) did not match the number "
-                f"of events found ({len(events)})"
+        if self.num_events is not None:
+            # Raise exception when events do not match
+            if not self.num_events == len(events):
```
istankovic, line 185: Please use `!=` rather than `not ... == ...`.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+    def __init__(
+        self, runner: scenario_runner.ScenarioRunner, config: Any, parent: "Task" = None
+    ) -> None:
+        config.update({"contract_name": "TokenNetwork", "event_name": "ChannelSettled"})
+        super().__init__(runner, config, parent)
+
+        verify_config(config, required_keys=["initiator", "partner", "channel_info_key"])
+
+        self.initiator = self._get_node_address(config["initiator"])
+        self.initiator_amount = config.get("initiator_amount", None)
```
istankovic, line 213: You can omit the second argument to `get`.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+        self.partner = self._get_node_address(config["partner"])
+        self.partner_amount = config.get("partner_amount", None)
```
istankovic, line 215: Same here.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+
+    def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-argument
+        channel_infos = self._runner.task_storage[STORAGE_KEY_CHANNEL_INFO].get(
+            self._config["channel_info_key"]
+        )
+        if channel_infos is None:
+            raise ScenarioError(
+                f"No stored channel info found for key '{self._config['channel_info_key']}'."
```
istankovic, line 223: Might be better to factor out `self._config['channel_info_key']` as a new variable, e.g. `key` and use that in the two places.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
             )
 
-        return {"events": events}
+        channel_identifier = channel_infos["channel_identifier"]
+        self.event_args.update(
+            {
+                "channel_identifier": channel_identifier,
+            }
+        )
```
istankovic, line 231: Shouldn't a simple `dict` setitem suffice here?
E.g. 
```python
self.event_args["channel_identifier"] = channel_identifier
```

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+            events = self._filter_for_channel_settled(
+                events, self.partner, self.partner_amount, self.initiator, self.initiator_amount
+            )
+        if len(events) != 1:
+            raise ScenarioAssertionError("Did not find expected ChannelSettled event!")
+        coop_settle_event = events[0]
+        transaction_hash = coop_settle_event["transactionHash"]
+        transaction = self.web3.eth.get_transaction(transaction_hash)
+        transaction_sender = transaction["from"]
+        if not transaction_sender == self.initiator:
```
istankovic, line 249: Please use `!=` rather than `not ... == ...`.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +173,97 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+    def _filter_for_channel_settled(
+        events, participant1, participant1_amount, participant2, participant2_amount
+    ):
+        event_args = {"participant1": participant1, "participant2": participant2}
+        if participant1_amount is not None:
+            event_args["participant1_amount"] = participant1_amount
+        if participant2_amount is not None:
+            event_args["participant2_amount"] = participant2_amount
+        event_args_items = event_args.items()
+        return [e for e in events if e["args"] and event_args_items & e["args"].items()]
```
istankovic, line 266: Do we need to return a mutable object from this function? Looking at the usage, I would say not, so I'd rather we return an immutable object, i.e. a tuple.

istankovic (changes requested):

scenario_player/tasks/blockchain.py
```diff
@@ -122,16 +133,10 @@ def __init__(
...
-        if not all_required_options_provided:
-            raise ScenarioError(
-                "Not all required keys provided. Required: " + ", ".join(required_keys)
-            )
-
+        verify_config(config, required_keys=["contract_name", "event_name"])
         self.contract_name = config["contract_name"]
         self.event_name = config["event_name"]
-        self.num_events = config["num_events"]
+        self.num_events = config.get("num_events")
```
ezdac, line 139: One thing I just noticed - if this is used in a scenario without the `num_events`, this will pass if there are no events, and also if there are events.
For internal use this makes sense, if one only wants to retrieve the events and not assert anything.

scenario_player/tasks/blockchain.py
```diff
@@ -24,6 +24,9 @@
 log = structlog.get_logger(__name__)
 
 
+CHANNEL_INFO_KEY = "channel_info_key"
```
istankovic, line 27: Does this need to be public?

scenario_player/tasks/blockchain.py
```diff
@@ -86,6 +89,14 @@ def query_blockchain_events(
     ]
 
 
+def verify_config(config, required_keys):
```
istankovic, line 92: Does this need to be public?

scenario_player/tasks/blockchain.py
```diff
@@ -86,6 +89,14 @@ def query_blockchain_events(
...
+    all_required_options_provided = all(key in config.keys() for key in required_keys)
+    if not all_required_options_provided:
```
istankovic, line 94: Nitpick: I think it'd be shorter/nicer to just do
```python
    if not all(key in config for key in required_keys):
    ...
```

scenario_player/tasks/blockchain.py
```diff
@@ -122,16 +133,10 @@ def __init__(
...
         super().__init__(runner, config, parent)
 
-        required_keys = ["contract_name", "event_name", "num_events"]
-        all_required_options_provided = all(key in config.keys() for key in required_keys)
-        if not all_required_options_provided:
-            raise ScenarioError(
-                "Not all required keys provided. Required: " + ", ".join(required_keys)
-            )
-
+        verify_config(config, required_keys=["contract_name", "event_name"])
```
istankovic, line 136: Nitpick: I'd rather use a tuple instead of a list here since we don't need mutability for `required_keys`.

scenario_player/tasks/blockchain.py
```diff
@@ -171,23 +176,94 @@ def _run(self, *args, **kwargs) -> Dict[str, Any]:  # pylint: disable=unused-arg
...
+    SYNCHRONIZATION_TIME_SECONDS = 0
+    DEFAULT_TIMEOUT = 5 * 60  # 5 minutes
+
+    def __init__(
+        self, runner: scenario_runner.ScenarioRunner, config: Any, parent: "Task" = None
+    ) -> None:
+        config.update({"contract_name": "TokenNetwork", "event_name": "ChannelSettled"})
+        super().__init__(runner, config, parent)
+
+        verify_config(config, required_keys=["initiator", "partner", CHANNEL_INFO_KEY])
```
istankovic, line 213: Nitpick: I'd rather use tuples instead of lists where mutability is not required.

palango:
A test would be nice. But I saw that there's none for `AssertBlockchainEventsTask` either.

scenario_player/tasks/blockchain.py
netcriptus: You can shorten this by stopping in the first key you don't find:

```python
if any(key not in config for key in required_keys):
  raise ScenarioError(
            "Not all required keys provided. Required: " + ", ".join(required_keys)
        )
```

PR #704, opened 2021-09-16
Add interactive wait task

konradkonrad:


weilbith (approved):
LGTM 🤣

ulope (dismissed):
This won’t work for multiple reasons:
- `input()` is blocking and will cause greenlets not to run while the prompt is displayed.
- When the UI is in use the prompt may not show up (although the ui could be used to solve this)

weilbith:
I'll fix the formatting with an amended commit and force push. I don't see the reason to first put a fixup commit here.

konradkonrad:
> This won’t work for multiple reasons:
> 
>     * `input()` is blocking and will cause greenlets not to run while the prompt is displayed.

While we're waiting for input, I don't see what kind of background greenlets need to run?
> 
>     * When the UI is in use the prompt may not show up (although the ui could be used to solve this)
I am aware of the UI issue -- the prompt does show up, but it does not look nice and ui rendering is broken afterwards.

ulope:
We clarified this in a call. This is a debugging only feature and therefore it's fine to merge as is.

PR #707, opened 2022-03-09
WIP: adapt scenario-player to run on arbitrum

netcriptus:
WIP

### DIFF
--- a/pfs1_get_a_simple_path.yaml
+++ b/pfs1_get_a_simple_path.yaml
@@ -1,0 +1,68 @@
+version: 2
+
+settings:
+  gas_price: {{ gas_price }}
+  services:
+    pfs:
+      url: {{ pfs_with_fee }}
+    udc:
+      enable: true
+      token:
+        deposit: true
+
+token:
+  address: "{{ transfer_token }}"
+  balance_fund: 1_000_000_000_000_000_000
+
+nodes:
+  count: 4
+  raiden_version: local
+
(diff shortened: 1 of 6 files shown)


pfs1_get_a_simple_path.yaml
```diff
@@ -0,0 +1,68 @@
+version: 2
```
andrevmatos, line 1: This file shouldn't be here

scenario_player/environment/arbitrum.json
```diff
@@ -0,0 +1,15 @@
...
+        "https://rinkeby.arbitrum.io/rpc"
+    ],
+    "transfer_token": "0x99C1aE38292791a52E26eE6AFdAeC69563F9b5C3",
+    "pfs_fee": 100,
+    "ms_reward_with_margin": 88000000000000000000,
+    "settlement_timeout_min": 40,
+    "raiden_client": "/Users/fernando/projects/light-client/raiden-cli/raiden",
+    "wait_short": 30,
+    "wait_long": 60,
+    "gas_price": "rpc"
```
andrevmatos, line 14: You'll probably need:
```suggestion
    "gas_price": "rpc",
    "matrix_servers": [ "auto", "auto", "auto", "auto" ]
```
`auto` let nodes detect its servers (which should work with LC); 4 of them will make it work with some scenarios which expect at least 4 servers to be specified

PR #709, opened 2022-08-04
Use raiden-common

karlb:
Use raiden-common instead of raiden as a dependency. This will allow us to depend on less and better maintained code as the python client gets deprecated.

Part of https://github.com/raiden-network/raiden/issues/7506

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,6 @@ commands:
             tar zxf geth.tar.gz
             cd geth*/
             install -m 755 geth ~/.local/bin/geth
-            pip install -U poetry
             curl -sSL https://install.python-poetry.org | python3 - --preview
             echo 'export PATH=~/.local/bin:$PATH' >> ${BASH_ENV}
             echo 'export LANG=C.UTF-8' >> ${BASH_ENV}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,7 @@ commands:
             cd geth*/
             install -m 755 geth ~/.local/bin/geth
             pip install -U poetry
+            curl -sSL https://install.python-poetry.org | python3 - --preview
             echo 'export PATH=~/.local/bin:$PATH' >> ${BASH_ENV}
             echo 'export LANG=C.UTF-8' >> ${BASH_ENV}
             echo 'export LC_ALL=C.UTF-8' >> ${BASH_ENV}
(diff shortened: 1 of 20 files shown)


scenario_player/environment/goerli-latest.json
```diff
@@ -0,0 +1,15 @@
...
+    "environment_type": "development",
+    "matrix_server_list": "https://raw.githubusercontent.com/raiden-network/raiden-service-bundle/goerli-latest-servers/known_servers/known_servers-goerli-latest.json",
+    "pfs_with_fee": "https://pfs.goerli-latest1.env.raiden.network",
+    "eth_rpc_endpoints": [
+        "http://geth.goerli.ethnodes.brainbot.com:8545"
+    ],
+    "transfer_token": "0xC563388e2e2fdD422166eD5E76971D11eD37A466",
+    "pfs_fee": 100,
+    "ms_reward_with_margin": 5500000000000000000,
+    "settlement_timeout_min": 40,
```
andrevmatos, line 11: Just FYI, the 0.50 `deployment_goerli.json` sets the `settle_timeout` to 15k seconds (~4h). This value here would need to match that, but either way, 4h is quite too long for settle_timeout on scenarios. We may want to redeploy those with reasonable testnet timeouts of 180s (e.g. as done in Arbitrum Rinkeby contracts). No need to worry now anyway

.circleci/config.yml
```diff
@@ -35,6 +35,7 @@ commands:
             cd geth*/
             install -m 755 geth ~/.local/bin/geth
             pip install -U poetry
```
andrevmatos, line 37: Is this `pip` call still required if we're going with the script way?

andrevmatos (approved):
lgtm, some small headsupl, but looks fine; this also needs to address the `arbitrum` branch for now, I think, while we still want to support mainnet on the SP servers

.circleci/config.yml
karlb: No it isn't. I removed that line.

karlb:
My goal would be to get the master branch working for both Arbitrum and normal Ethereum and to keep a branch with older code which can run on the 0.40 contracts (if we need that; I would prefer to shut those down sooner rather than later).

Should I merge this, or keep the PR open for now?

andrevmatos:
You can go ahead and merge it as is then. As long as master keeps working with contracts 0.40 while we still support that deployment on mainnet, this should be fine. Thank you

PR #710, opened 2022-08-11
Add scenarios to repo

karlb:
With the python client being deprecated, the scenarios need to find a
new place. This repo seems to be the best fit, since it allows a change
to touch the SP and the scenarios it works on in a single commit.

### DIFF
--- a/scenarios/bf1_basic_functionality.yaml
+++ b/scenarios/bf1_basic_functionality.yaml
@@ -1,0 +1,263 @@
+version: 2
+
+settings:
+  gas_price: {{ gas_price }}
+  services:
+    pfs:
+      url: "{{ pfs_with_fee }}"
+    udc:
+      enable: true
+      token:
+        deposit: true
+        balance_per_node: "{{ ms_reward_with_margin + 200 * 2 * pfs_fee}}"
+
+token:
+  address: "{{ transfer_token }}"
+  balance_fund: 10_000_000_000_000_000_000
+
+nodes:
+  count: 5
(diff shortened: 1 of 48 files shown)


scenarios/bf1_basic_functionality.yaml
```diff
@@ -0,0 +1,263 @@
...
+    default-reveal-timeout: {{ settlement_timeout_min // 4 }}
+  node_options:
+    0:
+      matrix-server: "{{ matrix_servers[0] }}"
+    1:
+      matrix-server: "{{ matrix_servers[1] }}"
+    2:
+      matrix-server: "{{ matrix_servers[2] }}"
+    3:
+      matrix-server: "{{ matrix_servers[3] }}"
```
andrevmatos, line 46: I think we have only 2 servers now, don't we? It'd be good to have this already adapted to work with the number of servers we have, so the SP server can already pick that up

andrevmatos (approved):
I agree moving the scenarios to SP repo is the best way forward. Even though there may be some differences between mainnet and arbitrum scenarios, they should also be possible to be split with the different SP branches (`master` vs `arbitrum` here). Besides the set of working matrix servers/PFSs, this LGTM, thanks

scenarios/bf1_basic_functionality.yaml
karlb: IIRC, the SP goes through the available number of servers multiple times if not enough servers are configured. If you actually got an error about it, let me know.

